PR #4, opened 2024-04-03
Issue 3: Not using DT.

### LINKED ISSUE #3: DT not exported from data.table
Two options... either we change to 

```
data.table:::DT
```

or move away from using data.table completely and switch to using dplyr (`mutate`?). 

Am happy to make these changes. Let me know what you think.

parksw3:
## Description

This PR closes #3 .

- I got rid of all instances of DT and replaced them with standard data.table uses. I tested that the vignette gives sensible answers.
- Also updated the vignette file to remove the use of DT.

## Checklist

- [x] I have added a news item linked to this PR.

### DIFF
--- a/R/fitting-and-postprocessing.R
+++ b/R/fitting-and-postprocessing.R
@@ -3,8 +3,7 @@
 sample_model <- function(model, data, scenario = data.table::data.table(id = 1),
                          diagnostics = TRUE, ...) {
 
-  out <- scenario |>
-    copy()
+  out <- data.table::copy(scenario)
 
   # Setup failure tolerant model fitting
   fit_model <- function(model, data, ...) {
@@ -16,12 +15,10 @@ sample_model <- function(model, data, scenario = data.table::data.table(id = 1),
   fit <- safe_fit_model(model, data, ...)
 
   if (!is.null(fit$error)) {
-    out <- out |>
-      DT(, error := list(fit$error[[1]]))
+    out[, error := list(fit$error[[1]])]
     diagnostics <- FALSE
   }else {
-    out <- out |>
-      DT(, fit := list(fit$result))
+    out[, fit := list(fit$result)]
     fit <- fit$result
   }
 
@@ -57,8 +54,7 @@ sample_epinowcast_model <- function(
   diagnostics = TRUE, ...
 ) {
 
-  out <- scenario |>
-    copy()
+  out <- data.table::copy(scenario)
 
   # Setup failure tolerant model fitting
   fit_model <- function(model, data, ...) {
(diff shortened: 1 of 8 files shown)


parksw3:
By the way, I don't have permission to commit directly to this repo...

parksw3:
It looks like this is failing because we have no tests set up...?

R/fitting-and-postprocessing.R
```diff
@@ -16,12 +15,10 @@ sample_model <- function(model, data, scenario = data.table::data.table(id = 1),
   fit <- safe_fit_model(model, data, ...)
 
   if (!is.null(fit$error)) {
-    out <- out |>
-      DT(, error := list(fit$error[[1]]))
+    out[, error := list(fit$error[[1]])]
```
seabbs, line 18: an annoying side effect of linting. Need to add error as a global var in utils to get rid of this flag.

seabbs: or we can update the linter to ignore global var checks (but then Rmd check will still flag this anyway)

R/fitting-and-postprocessing.R
```diff
@@ -123,11 +117,12 @@ sample_epinowcast_model <- function(
 #' Add natural scale summary parameters for a lognormal distribution
 #' @export
 add_natural_scale_mean_sd <- function(dt) {
-  nat_dt <- dt |>
-    data.table::DT(, mean := exp(meanlog + sdlog ^ 2 / 2)) |>
-    data.table::DT(,
-     sd := exp(meanlog + (1 / 2) * sdlog ^ 2) * sqrt(exp(sdlog ^ 2) - 1)
-    )
+  nat_dt <- data.table::copy(dt)
```
seabbs, line 120: in the original (i.e. above) we aren't making a copy but I think its fine either way (though add might imply no copy?)

R/fitting-and-postprocessing.R
```diff
@@ -207,10 +201,11 @@ extract_epinowcast_draws <- function(
...
 #' @export
 primary_censoring_bias_correction <- function(draws) {
-  draws <- data.table::copy(draws) |>
-    DT(, mean := mean - runif(.N, min = 0, max = 1)) |>
-    DT(, meanlog := log(mean^2 / sqrt(sd^2 + mean^2))) |>
-    DT(, sdlog := sqrt(log(1 + (sd^2 / mean^2))))
+  draws <- data.table::copy(draws)
+  draws[, mean := mean - runif(.N, min = 0, max = 1)]
+  draws[, meanlog := log(mean^2 / sqrt(sd^2 + mean^2))]
+  draw[, sdlog := sqrt(log(1 + (sd^2 / mean^2)))]
```
seabbs, line 207:
```suggestion
  draws[, sdlog := sqrt(log(1 + (sd^2 / mean^2)))]
```

R/fitting-and-postprocessing.R
seabbs: alternatively the env features we use elsewhere may help here?

seabbs (approved):
Nice ❤️! The linting flags are from global variables throwing problems and can be dealt with in another issue I think. 

We might also want an issue for making some of the `data.table` a bit cleaner without pipes as rather than doing each operation on a single line we might now want to start doing `dt[, `:=`(one_thing, two_thing, three_thing)]`.

Looking at the check now to see what the problem is.

seabbs:
> It looks like this is failing because we have no tests set up...?

Yes exactly. I have only partially added all the package infra this needs unfortunately.

PR #28, opened 2024-05-13
Issue 15: Remove excess models

### LINKED ISSUE #15: Which models to keep?
We currently have a range of models that have been ported over from the research code base. @parksw3 we need to decide which of these to keep prior to any work being done on these. 

I think my proposal would be keep just the latent truncation and censoring adjusted model and the dynamic backward -> forward models and with the view we will likely have a few more models flavours in the near future (as various improvements on those).

athowes:
## Description

This PR closes #15.

Remove all models but `latent_truncation_censoring_adjusted_delay()`.

## Checklist


athowes:
Remaining issues here I think are:

* [ ] Need to update documentation (e.g. do `document()`)
* [ ] Need to remove use of functions in vignette
* [ ] Perhaps there are functions which only are used in conjuction with a deprecated function?

athowes:
The vignette code probably doesn't work as expected at the moment, but seeing as we have other issues to fix that soon and it is on `eval = FALSE` I don't think that should stop us from merging this.

seabbs (approved):
Agree. Looks good.

seabbs:
> Perhaps there are functions which only are used in conjuction with a deprecated function?

Yes there are and this is partly where the discussion around having variables for events and delays comes in (as this model only users the former. I think (as you also appear to have felt) that this can be addressed in a later issue/PR.

PR #32, opened 2024-05-14
Issue 9: Getting started vignette

### LINKED ISSUE #9: Expand getting started vignette
The `epidist` [vignette](https://github.com/epinowcast/epidist/blob/main/vignettes/epidist.Rmd) is quite compact and possibly difficult to read for users not familiar with the package and methods already. I suggest expanding the writing to make it easier for new users to follow. The scope of this work would be to add further descriptions, links and references, figures, tables to the getting started vignette.

athowes:
## Description

This PR will close #9.
It is a draft of a rewritten version of the first "Get started" vignette.

## Checklist

### DIFF
--- a/vignettes/epidist.Rmd
+++ b/vignettes/epidist.Rmd
@@ -89,9 +89,16 @@ secondary_dist <- data.table(meanlog = 1.8, sdlog = 0.5) |>
 
 ```{r lognormal, fig.cap="(ref:lognormal)"}
 ggplot(data.frame(x = c(0, 8)), aes(x = x)) +
-  geom_function(fun = dlnorm, meanlog = secondary_dist[["meanlog"]], sdlog = secondary_dist[["sdlog"]]) +
+  geom_function(
+    fun = dlnorm,
+    meanlog = secondary_dist[["meanlog"]],
+    sdlog = secondary_dist[["sdlog"]]
+  ) +
   theme_minimal() +
-  labs(x = "Delay between primary and secondary event (days)", y = "Probability density")
+  labs(
+    x = "Delay between primary and secondary event (days)",
+    y = "Probability density"
+  )
 ```
 
 ```{r}
@@ -107,7 +114,10 @@ obs <- outbreak |>
 ```{r delay, fig.cap="(ref:delay)"}
 obs[case %% 50 == 0, ] |>
   ggplot(aes(y = case)) +
-  geom_segment(aes(x = ptime, xend = stime, y = case, yend = case), col = "grey") +
+  geom_segment(
+    aes(x = ptime, xend = stime, y = case, yend = case),
+    col = "grey"
+  ) +
   geom_point(aes(x = ptime), col = "#56B4E9") +
   geom_point(aes(x = stime), col = "#009E73") +
   labs(x = "Event time (day)", y = "Case number") +
@@ -164,7 +174,8 @@ sample_size <- 200
 This sample size corresponds to `r 100 * round(sample_size / nrow(obs_cens_trunc), 3)`% of the data.
 
 ```{r}
-obs_cens_trunc_samp <- obs_cens_trunc[sample(1:.N, sample_size, replace = FALSE)]
+obs_cens_trunc_samp <-
+  obs_cens_trunc[sample(seq_len(.N), sample_size, replace = FALSE)]
 ```
 
 With our censored, truncated, and sampled data, we are now ready to attempt to recover the underlying delay distribution using `epidist`.
--- a/vignettes/epidist.Rmd
+++ b/vignettes/epidist.Rmd
@@ -5,6 +5,7 @@ output:
   bookdown::html_vignette2:
     fig_caption: yes
     code_folding: show
+    number_sections: true
 pkgdown:
   as_is: true
 csl: https://raw.githubusercontent.com/citation-style-language/styles/master/apa-numeric-superscript-brackets.csl
@@ -38,8 +39,8 @@ library(purrr)
 library(ggplot2)
 ```
 
-Many important quantities in epidemiology can be thought of as the time between two events, or "delays".
-Examples include:
+Many quantities in epidemiology can be thought of as the time between two events, or "delays".
+Important examples include:
 
 * the incubation period (time between infection and symptom onset)
 * serial interval (time between symptom onset of infectee and symptom onset of infected), and 
@@ -47,23 +48,26 @@ Examples include:
 
 Unfortunately, estimating delays accurately from observational data is usually difficult.
 In our experience, the two main issues are interval censoring and right truncation.
-The `epidist` package makes it easy to estimate delays using a statistical model which properly accounts for these two issues.
+Don't worry if you've not come across these terms before.
+By simulating example data, we will explain what they mean in Section \@ref(data)!
+Next, in Section \@ref(fit), we show how `epidist` can be used to estimate delays using a statistical model which properly accounts for these two issues.
+Finally, in Section \@ref(compare), we demonstrate that the fitted delay distribution accurately recovers the underlying truth.
 
-## Example data
+# Example data {#data}
 
-Data must be in a particular format to be used within the `epidist` package.
-In this section, we use package functions to simulate data of the appropriate format.
+Data should be in a certain format for use within the `epidist` package.
+In this section, we simulate data of the appropriate format, and in doing so explain the two main issues with observational delay data.
 <!-- Please see as of yet unwritten vignette for information about how to transform your data of other formats to the right format. -->
 
-We start by using the [Gillepsie algorithm](https://en.wikipedia.org/wiki/Gillespie_algorithm) to generate infectious disease outbreak data from a stochastic compartmental model:
+First, we use the [Gillepsie algorithm](https://en.wikipedia.org/wiki/Gillespie_algorithm) to generate infectious disease outbreak data from a stochastic compartmental model:
 
 ```{r}
 outbreak <- simulate_gillespie(seed = 101)
 ```
 
-`outbreak` is a `data.table` object, with two columns: the case number `case` and the time of primary event `ptime`.
+`outbreak` is a `data.table` with columns for the case number `case` and the time of primary event `ptime`.
 
-For the delay between primary and secondary event, we will use a lognormal distribution:
+Now, to generate secondary events, we will use a lognormal distribution ror the delay between primary and secondary events:
 
 ```{r}
 secondary_dist <- data.table(meanlog = 1.8, sdlog = 0.5) |>
@@ -78,28 +82,61 @@ obs <- outbreak |>
 head(obs)
 ```
 
-Now we are ready to simulate observation of the epidemic process with both interval censoring and right truncation.
-Using `observe_process` we censor primary and secondary event times with a daily interval:
+`obs` is a `data.table` object with additional columns for the delay `delay` and time of secondary event `stime`.
+The secondary event time is simply the primary event time plus the delay:
 
 ```{r}
+all(obs$ptime + obs$delay == obs$stime)
+```
+
+If we received the data `obs` as above then estimating the delay distribution would be easy, and this package wouldn't need to exist.
+However, in reality, we will almost never receive the data as above.
+
+First, the times of primary and secondary events will usually be censored.
+This means that rather than exact event times, we observe event times within an interval.
+Here we suppose that the interval is daily, resulting in an obscuring of the exact delay data (Figure \@ref(fig:cens)):
+
+```{r}
+# observe_process() should be renamed and include a "daily" i.e. 1 argument
 obs_cens <- obs |> observe_process()
 ```
 
-Using `filter_obs_by_obs_time` we suppose that only individuals with a secondary event time before `obs_time` are observed:
+(ref:cens) Interval censoring of the primary and secondary event times obscures the delay times.
+
+```{r cens, fig.cap="(ref:cens)"}
+ggplot(obs_cens, aes(x = delay, y = delay_daily)) +
+  geom_point(alpha = 0.25) +
+  geom_abline(slope = 1, intercept = 0, linetype = "dashed") +
+  theme_minimal() +
+  coord_fixed() +
+  labs(x = "Exact delay time (days)", y = "Censored delay time (days)")
+```
+
+Next, during an outbreak we will usually be estimating delays in real time.
+The result is that only those cases with a secondary event occuring before some time will be observed.
+This is called (right) truncation, and biases the observation process towards shorter delays:
 
 ```{r}
 obs_time <- 25
-obs_cens_trunc <- filter_obs_by_obs_time(obs_cens_trunc, obs_time = obs_time)
+# filter_obs_by_obs_time() should be renamed to refer to stime
+obs_cens_trunc <- filter_obs_by_obs_time(obs_cens, obs_time = obs_time)
 ```
 
-Finally, we suppose that only a random sample of individuals of size `sample_size` are observed:
+Finally, in reality, it's not possible to observe every case. We suppose that a random sample of individuals of size `sample_size` are observed:
 
 ```{r}
 sample_size <- 200
+```
+
+This sample size corresponds to `r 100 * round(sample_size / nrow(obs_cens_trunc), 3)`% of the data.
+
+```{r}
 obs_cens_trunc_samp <- obs_cens_trunc[sample(1:.N, sample_size, replace = FALSE)]
 ```
 
-## Fit the model
+With our censored, truncated, and sampled data, we are now ready to attempt to recover the underlying delay distribution using `epidist`.
+
+# Fit the model {#fit}
 
 If we had access to the data `obs` it would be simple to estimate the delay distribution.
 However, with only access to the censored, truncated, sampled data `obs_cens_trunc_samp` it is not so easy...
@@ -111,7 +148,7 @@ fit <- latent_truncation_censoring_adjusted_delay(
 )
 ```
 
-## Compare estimates to reality
+# Compare estimates {#compare}
 
 Compare modelled estimate to underlying truth here to convince user that it works.
 Perhaps as an exercise, we could show the "really simple" estimate being wrong here otherwise it won't be so impressive that the model is right.
--- a/vignettes/epidist.Rmd
+++ b/vignettes/epidist.Rmd
@@ -78,34 +78,36 @@ obs <- outbreak |>
 head(obs)
 ```
 
-Now we are ready to simulate observation of the epidemic process at time `obs_time`:
-<!-- Note: Adam doesn't really understand `observe_process()` and `filter_obs_by_obs_time()` yet. -->
-<!-- When he does he'll write something better here. -->
+Now we are ready to simulate observation of the epidemic process with both interval censoring and right truncation.
+Using `observe_process` we censor primary and secondary event times with a daily interval:
+
+```{r}
+obs_cens <- obs |> observe_process()
+```
+
+Using `filter_obs_by_obs_time` we suppose that only individuals with a secondary event time before `obs_time` are observed:
 
 ```{r}
 obs_time <- 25
-
-obs <- obs |>
-  observe_process() |>
-  filter_obs_by_obs_time(obs_time = obs_time)
+obs_cens_trunc <- filter_obs_by_obs_time(obs_cens_trunc, obs_time = obs_time)
 ```
 
 Finally, we suppose that only a random sample of individuals of size `sample_size` are observed:
 
 ```{r}
 sample_size <- 200
-obs <- obs[sample(1:.N, sample_size, replace = FALSE)]
+obs_cens_trunc_samp <- obs_cens_trunc[sample(1:.N, sample_size, replace = FALSE)]
 ```
 
 ## Fit the model
 
-We fit the model adjusting for right truncation and date censoring using a latent variable approach.
-We briefly describe the model, and the inference methodology.
-Eventually, we link to a vignette about common issues with model fitting and HMC.
+If we had access to the data `obs` it would be simple to estimate the delay distribution.
+However, with only access to the censored, truncated, sampled data `obs_cens_trunc_samp` it is not so easy...
+Fortunately we solve it!
 
 ```{r}
 fit <- latent_truncation_censoring_adjusted_delay(
-  data = obs, cores = 4, refresh = 0
+  data = obs_cens_trunc_samp, cores = 4, refresh = 0
 )
 ```
 
--- a/vignettes/epidist.Rmd
+++ b/vignettes/epidist.Rmd
@@ -96,11 +96,13 @@ secondary_dist <- data.table(meanlog = 1.8, sdlog = 0.5) |>
 (ref:lognormal) The lognormal distribution is skewed to the right. Long delay times still have some probability.
 
 ```{r lognormal, fig.cap="(ref:lognormal)"}
-ggplot(data.frame(x = c(0, 8)), aes(x = x)) +
+ggplot(data.frame(x = c(0, 30)), aes(x = x)) +
   geom_function(
     fun = dlnorm,
-    meanlog = secondary_dist[["meanlog"]],
-    sdlog = secondary_dist[["sdlog"]]
+    args = list(
+      meanlog = secondary_dist[["meanlog"]],
+      sdlog = secondary_dist[["sdlog"]]
+    )
   ) +
   theme_minimal() +
   labs(
@@ -189,7 +191,7 @@ obs_cens_trunc_samp <-
 
 With our censored, truncated, and sampled data, we are now ready to try to recover the underlying delay distribution using `epidist`.
 
-# Fit the model {#fit}
+# Fit the model and compare estimates {#fit}
 
 If we had access to the data `obs`, then it would be simple to estimate the delay distribution.
 However, with only access to `obs_cens_trunc_samp`, the delay distribution we observe is biased (Figure \@ref(fig:obs-est)) and we must use a statistical model.
@@ -215,13 +217,15 @@ dplyr::bind_rows(
     data = data.frame(x = c(0, 30)),
     aes(x = x),
     fun = dlnorm,
-    meanlog = secondary_dist[["meanlog"]],
-    sdlog = secondary_dist[["sdlog"]],
+    args = list(
+      meanlog = secondary_dist[["meanlog"]],
+      sdlog = secondary_dist[["sdlog"]]
+    ),
   ) +
   labs(
     x = "Delay between primary and secondary event (days)",
     y = "Probability density",
-    type = ""
+    fill = ""
   ) +
   theme_minimal() +
   theme(legend.position = "bottom")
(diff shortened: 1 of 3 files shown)


seabbs:
This looks good so far. Give me a ping when you want a proper review

athowes:
@seabbs @kgostic -- I'm not fully done with this but would be interested in some review / feedback to make sure that I'm staying on the right path.

If you'd prefer to wait until the vignette is more fully complete, then let me know.

athowes:
I will now try fixing this `pkgdown` error so that you should hopefully be able to extract out the `.html` version of the vignette.

athowes:
I don't know what is causing the issue with the `pkgdown` website on check. I can render it OK locally. Let me know if you have any idea. It is also there on the other PR so it must be something already in `main`.

kgostic:
@seabbs -- can you help with this tomorrow?

vignettes/epidist.Rmd
```diff
@@ -1,201 +1,211 @@
...
-Simulate data from an outbreak.
+1. interval censoring, and
+2. right truncation.
 
-```{r simulate-outbreak}
+Don't worry if you've not come across these terms before!
+In Section \@ref(data), we will explain what they mean by simulating example data analogous to what we might observe during an ongoing infectious disease outbreak.
+Next, in Section \@ref(fit), we show how `epidist` can be used to estimate delays using a statistical model which properly accounts for these two issues.
+Finally, in Section \@ref(compare), we demonstrate that the fitted delay distribution accurately recovers the underlying truth.
+
```
kgostic, line 56:
```suggestion

```suggestion

If you want more technical details: the `epidist` package implements models following best practices described in [Park et al. 2023](https://www.medrxiv.org/content/10.1101/2024.01.12.24301247v1) and [Charniga et al. 2024](https://arxiv.org/abs/2405.08841).
```
```

File: vignettes/epidist.Rmd [hunk 2]
```diff
@@ -1,201 +1,211 @@
...
+# Example data {#data}
+
+Data should be in a certain format for use within the `epidist` package.
```
kgostic, line 59: Here, I'd insert a few lines describing the required data schema, including the required fields, a definition for each field, and notes on data type requirements, e.g.:

```suggestion
Data should be formatted as a [`data.table`]() with the following columns for use within the `epidist` package:

* `case` (data type): a unique ID for each observed case
* `ptime` (data type): ...
* etc. 
```

vignettes/epidist.Rmd
```diff
@@ -1,201 +1,211 @@
...
-    sdlog = secondary_dist$sdlog[[1]]
-  ) |>
-  observe_process()
+    meanlog = secondary_dist[["meanlog"]],
+    sdlog = secondary_dist[["sdlog"]]
+  )
 ```
 
-Observe the outbreak after 25 days and take 100 samples.
+(ref:delay) Secondary events (in green) occur at random  (As with Figure \@ref(fig:outbreak), to make this figure easier to read, only every 50th case is shown.)
```
kgostic, line 112: I think many non-statistical users see "at random" and think "unpredictably", but you're really trying to convey that this is a stochastic process. Consider rewording?

vignettes/epidist.Rmd
```diff
@@ -1,201 +1,211 @@
...
+If we received the data `obs` as above then estimating the delay distribution would be easy, and this package wouldn't need to exist.
+However, in reality, we will almost never receive the data as above.
 
-```{r cohort-observed-cases}
-truncated_cases <- construct_cases_by_obs_window(
-  obs, windows = c(25), obs_type = "ptime"
-)
+First, the times of primary and secondary events will usually be censored.
+This means that rather than exact event times, we observe event times within an interval.
+Here we suppose that the interval is daily, resulting in an obscuring of the exact delay data (Figure \@ref(fig:cens)):
```
kgostic, line 139:
```suggestion
Here we suppose that the interval is daily, meaning that only the date of the primary or secondary event, not the exact time of each event, is reported.  (Figure \@ref(fig:cens)):
```

vignettes/epidist.Rmd
```diff
@@ -1,201 +1,211 @@
...
+ggplot(obs_cens, aes(x = delay, y = delay_daily)) +
+  geom_point(col = "#E69F00") +
+  geom_abline(slope = 1, intercept = 0, linetype = "dashed", col = "grey") +
+  theme_minimal() +
+  coord_fixed() +
+  labs(x = "Exact delay time (days)", y = "Censored delay time (days)")
 ```
 
-Plot the mean difference between continuous and discrete event time:
+Next, during an outbreak we will usually be estimating delays in real time.
```
kgostic, line 157: Epidemiological delays are often ascertained via the secondary event (e.g. when estimating an incubation period the symptom onset event is much more likely to be observed and reported than the infection event).

vignettes/epidist.Rmd
```diff
@@ -1,201 +1,211 @@
...
-  facet_wrap(vars(parameter), nrow = 1, scales = "free_x") +
-  scale_fill_brewer(palette = "Dark2") +
-  guides(fill = guide_none()) +
-  labs(
-    y = "Model", x = "Relative to ground truth"
-  )
-```
+Compare modelled estimate to underlying truth here to convince user that it works.
+Perhaps as an exercise, we could show the "really simple" estimate being wrong here otherwise it won't be so impressive that the model is right.
+This section should also show the user how to get objects that they might want out of the fitted object.
```
kgostic, line 205: I think this could be nice, but raises questions about whether we should keep the "naive" model or just code it using brms in the example.

We should also choose a pretty short example delay interval if we want the adjusted estimate to be much better than the raw one.

kgostic (approved):
@athowes this looks great overall. Your writing is really clear and concise.

I do think there's a bit at the end that still needs to be fleshed out, but if you can finish off the end, I think this is a reasonable outline for a getting started vignette.

@seabbs may want to add a bit more about installation and contributing issues at the top.

I'm approving so as not to hold things up but it would be really nice to see a rendered version of the makrdown as I haven't yet seen the figures in situ.

vignettes/epidist.Rmd
athowes: Agree. Added with Bibtex. Arguably you might like the link?

vignettes/epidist.Rmd
athowes: "occur with a delay drawn from the lognormal distribution" --> chance to reinforce the delay distribution

vignettes/epidist.Rmd
athowes: Agree.

vignettes/epidist.Rmd
athowes: I think we should code it using `brms` in the example. Perhaps you could say it's the wrong place, but if someone is familiar with `y ~ x` synatx it might reinforce to them that our model is really just some fancy version of that.

Good point about choosing the simulation settings to make sure there is a difference. When I get to this I'll choose the simulation settings to demonstrate the value of using `epidist`.

seabbs:
Can you rebase this so the `pkgdown` builds and then I will review (I thought about rebasing myself but it seemed a bit cheeky)

vignettes/epidist.Rmd
athowes: I think this is a good suggestion. My only uncertainty is that I don't exactly know which columns are "hard requirements" for using the package and which are "there sometimes".

I suppose doing this will clarify this point for me (which is needed to do anyway).

seabbs:
> gree on making sure people have the package installed, but would be tempted to have that displayed very prominently in the README and not in the "getting started" vignette. 

Agree

vignettes/epidist.Rmd
seabbs: I think either or works. Personally I find links more useful but I get the portability arg behind just giving a citation

vignettes/epidist.Rmd
seabbs: Good point. I like the reword suggestions

vignettes/epidist.Rmd
seabbs: > **What happens if we had the symptom onset date for an individual but not the infection date? Discard their data? Then it's a non-representative sample?

Yes exactly (so no the methods here are not accounting for this) - `epinowcast` has a module for this (the missingness module) but its a tricky problem)

vignettes/epidist.Rmd
seabbs: > We should also choose a pretty short example delay interval

Or long distribution?

> I think we should code it using brms in the example.

I think I agree but we may look at this and change our mind if it seems a bit complicated

vignettes/epidist.Rmd
```diff
@@ -1,201 +1,250 @@
 ---
-title: "Getting Started with Epidist"
-description: "A quick start guide to using the Epidist package"
-author: Epidist Team
+title: "Getting started with epidist"
+description: "A quick start guide to using the epidist R package"
 output: 
-  bookdown::html_vignette2:
+  bookdown::html_document2:
```
seabbs, line 5: Can you explain?

vignettes/epidist.Rmd
```diff
@@ -1,201 +1,250 @@
...
     fig_caption: yes
     code_folding: show
+    number_sections: true
 pkgdown:
   as_is: true
-csl: https://raw.githubusercontent.com/citation-style-language/styles/master/apa-numeric-superscript-brackets.csl
+# csl: https://raw.githubusercontent.com/citation-style-language/styles/master/apa-numeric-superscript-brackets.csl
```
seabbs, line 11: ?

vignettes/epidist.Rmd
```diff
@@ -1,201 +1,250 @@
...
 )
 ```
 
-# Quick start
-
-In this quick start, we demonstrate using `epidist` to ...
-
-# Package
-
-In this quick start, we also use `data.table`, `purrr`, and `ggplot2` packages. These are both installed as dependencies when `epidist` is installed. Note that all output from `epidist` is readily useable with  other tools, including `tidyverse` packages (see [here](https://mgimond.github.io/rug_2019_12/Index.html) for a comparison).
```
seabbs, line 36: Why cut?

vignettes/epidist.Rmd
```diff
@@ -1,201 +1,250 @@
...
+  comment = "#>",
+  message = FALSE,
+  warning = FALSE,
+  error = FALSE
 )
 ```
 
-# Quick start
-
-In this quick start, we demonstrate using `epidist` to ...
```
seabbs, line 32: I quite like a summary - did you drop because you will do later or you don't think its a good idea?

vignettes/epidist.Rmd
```diff
@@ -1,201 +1,250 @@
...
+  theme_minimal()
+```
+
+`outbreak` is a [`data.table`](https://cran.r-project.org/web/packages/data.table/index.html) with the columns `case` and `ptime`.
+
+Now, to generate secondary events, we will use a lognormal distribution (Figure \@ref(fig:lognormal)) for the delay between primary and secondary events:
+
+```{r}
+secondary_dist <- data.table(meanlog = 1.8, sdlog = 0.5) |>
   add_natural_scale_mean_sd()
```
seabbs, line 93: Note there may be some work spinning up on making a small package to carry these transforms around

vignettes/epidist.Rmd
```diff
@@ -1,201 +1,250 @@
...
+(ref:delay) Secondary events (in green) occur with a delay drawn from the lognormal distribution (Figure \@ref(fig:lognormal)). As with Figure \@ref(fig:outbreak), to make this figure easier to read, only every 50th case is shown.
+
+```{r delay, fig.cap="(ref:delay)"}
+obs[case %% 50 == 0, ] |>
+  ggplot(aes(y = case)) +
+  geom_segment(
+    aes(x = ptime, xend = stime, y = case, yend = case),
+    col = "grey"
+  ) +
+  geom_point(aes(x = ptime), col = "#56B4E9") +
```
seabbs, line 129: I would usually do this by going to long form data and then using the new variable as then you can have a legend. This is nice and clear as is though

vignettes/epidist.Rmd
```diff
@@ -1,201 +1,250 @@
...
+obs_cens <- obs |> observe_process()
 ```
 
-We can alternatively plot the observed data based on primary events. This corresponds to a retrospective cohort based view of the data.
-
-```{r cohort-observed-cases}
-truncated_cases <- construct_cases_by_obs_window(
-  obs, windows = c(25), obs_type = "ptime"
-)
+(ref:cens) Interval censoring of the primary and secondary event times obscures the delay times.
```
seabbs, line 154:
```suggestion
(ref:cens) Interval censoring of the primary and secondary event times obscures the delay times. Note that while daily censoring is most common, both the primary and secondary events can have other delay intervals, and this scenario is supported by  `epidist`.
```

vignettes/epidist.Rmd
```diff
@@ -1,201 +1,250 @@
...
+  theme_minimal() +
+  coord_fixed() +
+  labs(x = "Exact delay time (days)", y = "Censored delay time (days)")
 ```
 
-Plot the true continuous delay distribution and the empirical observed distribution for each observation window.
+Next, during an outbreak we will usually be estimating delays in real time.
+The result is that only those cases with a secondary event occurring before some time will be observed.
+This is called (right) truncation, and biases the observation process towards shorter delays:
+<!-- I think possible to do a better job with showing what right truncation is with a figure here! -->
```
seabbs, line 168: yes good idea. We could reproduce one from one of the papers or role a new one. It feels like we could promote this to a issue vs doing here if we think it needs more thought.

In terms of what is in `epidist` I quite like the primary event bar plot with the dots for secondary events but can also see the clear case for it to be done on individual delays when explaining

vignettes/epidist.Rmd
```diff
@@ -1,201 +1,250 @@
...
+# filter_obs_by_obs_time() should be renamed to refer to stime
+obs_cens_trunc <- filter_obs_by_obs_time(obs_cens, obs_time = obs_time)
+```
 
-plot_empirical_delay(
-  combined_obs, meanlog = secondary_dist$meanlog[[1]],
-  sdlog = secondary_dist$sdlog[[1]]
-)
+Finally, in reality, it's not possible to observe every case.
+We suppose that a sample of individuals of size `sample_size` are observed:
```
seabbs, line 177: if we wanted to make a point about ascertainment here seems like it could be a good choice

vignettes/epidist.Rmd
```diff
@@ -1,201 +1,250 @@
...
-Adjust for right truncation and date censoring using a latent variable approach.
+# Fit the model {#fit}
 
-```{r}
-latent_truncation_censoring_fit <- latent_truncation_censoring_adjusted_delay(
-  data = truncated_obs, cores = 4, refresh = 0
-)
+If we had access to the data `obs`, then it would be simple to estimate the delay distribution (Figure \@ref(fig:obs-est)).
+
+(ref:obs-est) The histogram of delays matches closely the underlying distribution. Edit: at the moment something is wrong here.
```
seabbs, line 196: ruh roh. On the face of it I can't see a code issue. Can dig in if helpful

vignettes/epidist.Rmd
```diff
@@ -1,201 +1,250 @@
...
+    fun = dlnorm,
+    meanlog = secondary_dist[["meanlog"]],
+    sdlog = secondary_dist[["sdlog"]],
+  ) +
+  labs(x = "Delay between primary and secondary event (days)", y = "Probability density") +
+  theme_minimal()
 ```
 
-### Summarise model posterior and compare to known truth
+However, with only access to `obs_cens_trunc_samp`, we must use a statistical model to avoid severely biased estimates!
```
seabbs, line 212: Prove it to me with a plot

vignettes/epidist.Rmd
```diff
@@ -1,201 +1,250 @@
...
+Many of these are so-called "latent variables".
 
-```{r lognormal-draws}
-draws <- models |>
-  map(extract_lognormal_draws) |>
-  rbindlist(idcol = "model") |>
-  rbind(epinowcast_draws, use.names = TRUE)
+```{r pars}
+pars <- fit$fit@par_dims |>
+  map(.f = function(x) if(identical(x, integer(0))) return(1) else return(x))
```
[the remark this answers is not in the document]
seabbs: There are other choices that provide more functionality but are heavier like `tidybayes` (heavier in the sense of them themselves having more dependencies).

vignettes/epidist.Rmd
```diff
@@ -1,201 +1,250 @@
...
 
-draws <- draws[,
-   model := factor(
-    model, levels = c("Joint incidence and forward delay", rev(names(models)))
-   )]
+data.frame("Parameter" = names(pars), "Length" = unlist(pars)) |>
+  gt::gt(caption = "All of the parameters that are included in the model")
```
seabbs, line 233: As we don't have this as a dep I think we would need to add it as a suggest which I think is a valid thing to do

vignettes/epidist.Rmd
```diff
@@ -1,201 +1,250 @@
...
+```
 
-summarised_draws <- draws |>
-  draws_to_long() |>
-  summarise_draws(sf = 3)
+Although some users may wish to work with `fit` directly, we will see in the following section that `epidist` provides functions to make common post-processing tasks easy.
```
seabbs, line 236:
```suggestion
Although some users may wish to work with fit directly, we will see in the following section that epidist provides functions to make common post-processing tasks easy. Additionally, users can leverage tools that support brms models for fit objects.
```

My suggestion seems a bit clunky but I more generally think we need to flag this info

seabbs: Also I see from this seciton why you went with manual extraction in the previous section. I like it from a teaching sense my only concern would be does it add more verbosity (answer don't know what do you think?)

vignettes/epidist.Rmd
```diff
@@ -1,201 +1,250 @@
...
 
-knitr::kable(summarised_draws[parameter %in% c("meanlog", "sdlog")])
-```
+# Compare estimates {#compare}
 
-Plot summarised posterior estimates from each model compared to the ground truth.
+Compare modelled estimate to underlying truth here to convince user that it works.
+Perhaps as an exercise, we could show the "really simple" estimate being wrong here otherwise it won't be so impressive that the model is right.
```
seabbs, line 241: we could either show the simple model is wrong or we could show the truncated dist above more clearly vs the underlying dist? That might do a lot of what fitting a simple model would do?

vignettes/epidist.Rmd
seabbs: See above comment about maybe doing this via visualing the true dist and the truncated and censored dist. I think do that and then see if we need this? We might still as we can clearly show the summary parameters of the fits vs true (like the old readme but simpler).

seabbs (approved):
I really like this @athowes it is really clearly written and communicates the points very well. 

I put a few minor comments. My only other concern here is maybe length but uncertain.

vignettes/epidist.Rmd
athowes: I find that the `html_vignette2` format creates figures which are formatted strangely (low resolution, incorrect scale of different objects in the figure). I guess this is somewhat of a hack and there is a reason that we would want the vignette to be in `html_vignette2` format rather than `html_document2` format. If there is some way to make the figures less bad in `html_vignette2` please let me know.

vignettes/epidist.Rmd
athowes: This was so that "as described in Park et al. ([2024](http://127.0.0.1:36043/rmd_output/0/#ref-park2024estimating)) and Charniga et al. ([2024](http://127.0.0.1:36043/rmd_output/0/#ref-charniga2024best))." appears rather than e.g. [1], [2].

Ideally perhaps the `.bibtex` format would also include linking to the paper. I don't know if any of the `citation-style-language/` ones allow you to do that?

vignettes/epidist.Rmd
athowes: I don't think this is crucial information for someone reading the "get started" to be presented with at the beginning. If someone cares about dependencies then likely they know that they can look at the `DESCRIPTION` file to figure out the dependencies.

The point about output being readily usable with other tools seems somewhat obvious to me. Are people writing packages where the output is only usable with `data.table` say?

vignettes/epidist.Rmd
athowes: I would also usually do that but in some sense I was trying to make the number of lines used low.

vignettes/epidist.Rmd
athowes: Added.

vignettes/epidist.Rmd
seabbs: Okay fair enough. I don't have a strong preference here. I'm not clear if there is a reason for `html_vignette` apart from backwards compatible formatting and if its just that no issue not using

vignettes/epidist.Rmd
seabbs: Got you. Happy to drop the csl then

vignettes/epidist.Rmd
athowes: I'm also confused about why this is wrong.

vignettes/epidist.Rmd
seabbs: > If someone cares about dependencies then likely they know that they can look at the DESCRIPTION file to figure out the dependencies.

Ah so this is because these aren't always going to be package dependencies. We might (and do here - looking at you `gt`) use packages that only appear in the suggests. This means that when people follow our install guides they won't always be able to run the vignette without either installing the suggests or manually knowing about them. 

> The point about output being readily usable with other tools seems somewhat obvious to me. Are people writing packages where the output is only usable with data.table say?

I think it might be obvious to a lot of people but I also think this could scare quite a large minority of people so at the cost of the single sentence seems worth keeping.

vignettes/epidist.Rmd
athowes: Agree and done.

athowes: Still IDK what's going on with the true density..

image

vignettes/epidist.Rmd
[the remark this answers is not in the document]
athowes: I just didn't know an easier way to do this. Open to moving to `brms` / `posterior` / `tidybayes`. This is probably a one off thing for the vignette rather than a function in the package so not sure how much it matters.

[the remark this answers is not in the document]
athowes: We already have `posterior` in `Imports:`

vignettes/epidist.Rmd
athowes: Yeah I think we will use `gt` again in other vignettes so I'll add it.

vignettes/epidist.Rmd
athowes: Tried:

In the following section, we show that `epidist` provides functions to make common post-processing tasks easy. Additionally, for users familiar with Stan and `brms`, they may like to work with `fit` directly. For example, any tool that supports `brms` fitted model objects will be compatible with `fit`.

vignettes/epidist.Rmd
athowes: Yeah having done the above I agree, and "fitting a model" to the obscured data isn't really doing much. It does do some things like teach people we are not doing anything really magic, just an altered `brms` model, but unclear if it's worth it.

vignettes/epidist.Rmd
athowes: Worth mentioning in this vignette that we don't deal with it then? Or leave that to something more advanced? I'd suggest leave to more advanced.

vignettes/epidist.Rmd
seabbs: Hmm so the version we have here: https://github.com/parksw3/epidist-paper

definitely works as intended (we should talk a bit about why we are plotting from scratch here (is this just for now or is this a long term choice you have made).

seabbs: ah you aren't passing via an args list: https://github.com/parksw3/epidist-paper/blob/d34a461c4d7526438d509eec99025abda30687c7/R/plot.R#L68

do the parameters you are passing influence what is being plot?

vignettes/epidist.Rmd
seabbs: I think very briefly pointing it out (and at some later date linking to something more complex) might be a good user journey

vignettes/epidist.Rmd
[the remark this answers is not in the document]
seabbs: I think using poosterior might be a good idea. What you have right now is a bit scary

vignettes/epidist.Rmd
seabbs: yeeah agree - also for our impact factor we need to be magical ;)

seabbs:
When ready for another review cna you reping using the interface at top right (as this gives a different form of notification that I filter for vs saying something like ready for review)

vignettes/epidist.Rmd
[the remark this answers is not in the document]
athowes: Will try doing. I don't know the function in `posterior` to do this off the top of my head or if there is one.

vignettes/epidist.Rmd
athowes: Plotting from scratch because it's easier for me to write, rather than it being a long term thought out choice or something like that.

And yes that fixed it, great spot, thanks!

vignettes/epidist.Rmd
athowes: Yep!

vignettes/epidist.Rmd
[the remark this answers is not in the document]
athowes: I didn't find an easy way to do this with `posterior`.

vignettes/epidist.Rmd
athowes: Happy to leave to you and Katie to decide on this. I'd imagine putting ascertainment into a more complex vignette could be better (especially if there is any analysis the package can do to help you with it.).

athowes:
Great, thanks all for the comments both.

I'm going to give this another hour or so try, and then I think we should clear it up and merge it in to get things moving. Happy to come back and iterate it again in future.

Have enjoyed writing this and the comments / feedback process by the way! Very smooth / logistically easy to handle.

vignettes/epidist.Rmd
seabbs: I think we should make a concise point here and link out to something more complex inn the future

vignettes/epidist.Rmd
athowes: I still think this is a good suggestion but would like to leave to a later issue as I am unsure of the input requirements we want, and also unsure of the columns we want.

vignettes/epidist.Rmd
athowes: I've tried something brief about this. Hard to know where exactly to put it.

vignettes/epidist.Rmd
athowes: I think I've done this.

vignettes/epidist.Rmd
athowes: Added:

Finally, to run this vignette yourself, you will need the `data.table`, `purrr` and `ggplot2` packages installed.
Note that to work with outputs from `epidist` you do not need to use `data.table`: any tool of your preference is suitable.

vignettes/epidist.Rmd
athowes: Going to make an issue to address this.

seabbs (approved):
LGTM. This looks to have covered all review points from myself and @kgostic.

PR #58, opened 2024-05-24
Issue 53: Update simulate_gillespie() documentation and argument names

### LINKED ISSUE #53: Improve documentation for simulate_gillespie() to better explain parameterisation
Given that we have parameters `r` and `gamma` defined as
```R
#' @param r The rate of infection. Defaults to 0.2.
#' @param gamma The rate of recovery. Defaults to 1/7.
```
is the use of `beta` in the following lines muddled/erroneous?
https://github.com/epinowcast/epidist/blob/03cac6f7dc83b05ca6860965edbcd62175b6d449/R/simulate.R#L88-L93

If so I believe this would also effect the paper code at https://github.com/parksw3/epidist-paper/blob/main/R/simulate.R (but probably tangential to the results discussed).

Apologies for the noise if I'm mistaken.

athowes:
## Description

This PR closes #53.

* I have updated the documentation inline with the referenced issue and clarification made by @parksw3
* I have also changed the function arguments to be inline with the writing. That is, using `I0` and `N` rather than `init_I` and `n` in the function
* I have rerendered the `.md` for this function

## Checklist

### DIFF
--- a/R/simulate.R
+++ b/R/simulate.R
@@ -85,7 +85,7 @@ simulate_gillespie <- function(r = 0.2,
     set.seed(seed)
   }
   t <- 0
-  state <- c(n - I0, I0, 0)
+  state <- c(N - I0, I0, 0)
   beta <- r + gamma
   go <- TRUE
   ptime <- NULL
--- a/R/simulate.R
+++ b/R/simulate.R
@@ -58,9 +58,9 @@ simulate_exponential_cases <- function(r = 0.2,
   return(cases)
 }
 
-#' Simulate cases from a Stochastic SIR model
+#' Simulate cases from a stochastic SIR model
 #'
-#' This function simulates cases from an Stochastic SIR model. The user may
+#' This function simulates cases from an stochastic SIR model. The user may
 #' specify the initial epidemic growth rate \eqn{r}, the rate of recovery gamma
 #' \eqn{\gamma}, the initial number of infected cases \eqn{I_0}, and the total
 #' population size \eqn{N}.
--- a/R/simulate.R
+++ b/R/simulate.R
@@ -7,7 +7,7 @@
 #' @param t Upper bound of the uniform distribution to generate primary event
 #' times.
 #'
-#' @return A data table with two columns: case (case number) and ptime (primar
+#' @return A data table with two columns: case (case number) and ptime (primary
 #' event time).
 #'
 #' @family simulate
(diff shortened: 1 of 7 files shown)


seabbs:
This looks good to me but it would be nice to address the linting issues here just so they pass (even though some are unrelated to changes). 

I think I should leave the main review to @parksw3 but if they have no time I am also happy to be pinged to come back and do

athowes:
Thanks for the suggestion. Linting now fixed.

parksw3:
Realistically, I can review on Thursday if that's not too late. Have graduation today and tomorrow, and moving to Canada on Wednesday.

parksw3:
Actually, I just took a quick look at it (thought there would be a lot more to review so I suggested Thursday, sorry).

- For simulating exponential cases, I wonder if it makes sense to say that r is the growth rate, rather than the rate parameter. We're not quite simulating an exponential distribution because it's bounded.
- Not sure how to link code but L139: `delay <- ptime <- stime <- NULL` I'm not sure why this line is necessary in `simulate_secondary`?

R/simulate.R
```diff
@@ -134,18 +136,19 @@ simulate_gillespie <- function(r = 0.2,
 #' @family simulate
 #' @export
 simulate_secondary <- function(linelist, dist = rlnorm, ...) {
+  delay <- ptime <- stime <- NULL
```
parksw3, line 139: I'm not sure why this is necessary.

R/simulate.R
```diff
@@ -26,9 +26,10 @@ simulate_uniform_cases <- function(sample_size = 1000, t = 60) {
 #' If the rate parameter is 0, then this function defaults to the uniform
 #' distribution.
 #'
-#' @param r The rate parameter for the exponential distribution. Defaults to 0.2.
+#' @param r The rate parameter for the exponential distribution. Defaults to
```
parksw3, line 29: Is it clearer/more accurate to describe this as an exponential growth rate and avoid saying exponential distribution?

parksw3 (commented):
Left two comments.

R/simulate.R
athowes: Yes, it's silly, but done to prevent linting problems with `data.table`. See: https://cran.r-project.org/web/packages/data.table/vignettes/datatable-importing.html (Dealing with "undefined global functions or variables")

R/simulate.R
athowes: Maybe, though perhaps out of scope for this issue? (I've not changed this text aside from the spacing.)

seabbs: I agree w/ promoting to a new issue

R/simulate.R
seabbs: (its not just `data.table` its R package checks more generally)

seabbs: An alternative is to define them everywhere as global variables in utils which can be a bit less messy.

seabbs (dismissed):
This all looks good. I think we should either address here (as small) or make a new issue for the rewording and also make an issue for how to handle global variable warnings package wide. I have a small preference for using the global variables functionality vs defining to NULLS

athowes:
> I think we should either address here (as small)

I have done this and also addressed other small style issues with the `simulate.R` roxygen as I see it.

> also make an issue for how to handle global variable warnings package wide. I have a small preference for using the global variables functionality vs defining to NULLS

Doing this now.

seabbs (approved):

PR #59, opened 2024-05-24
Issue 57: Polish get started vignette

### LINKED ISSUE #57: Polish getting started vignette
Good that we merged in the "get started" vignette and now can see it rendered: http://epidist.epinowcast.org/articles/epidist.html

There are a few small issues I can see with it that require polishing up:

* [x] Move loading packages to where the loading packages is mentioned
* [ ] Alter the "Finally, in Section [??](http://epidist.epinowcast.org/articles/epidist.html#compare), we demonstrate that the fitted delay distribution accurately recovers the underlying truth." (I merged two sections)
* [ ] Ideally the Table 2.1 caption should be altered to work like figure captions rather than this strange thing. I feel like this is possible but requires trying a few things out. Basically I'd like to be using `(ref:...)`
* [ ] The section at the end still has "Figure [2.2](http://epidist.epinowcast.org/articles/epidist.html#fig:fitted-lognormal) shows…" and an empty figure caption that need to be filled out

Can make a quick PR to fix these things tomorrow.

athowes:
## Description

This PR closes #57.

* [x] Move loading packages to where the loading packages is mentioned
* [x] Alter the "Finally, in Section [??](http://epidist.epinowcast.org/articles/epidist.html#compare), we demonstrate that the fitted delay distribution accurately recovers the underlying truth." (I merged two sections)
* [x] Ideally the Table 2.1 caption should be altered to work like figure captions rather than this strange thing. I feel like this is possible but requires trying a few things out. Basically I'd like to be using `(ref:...)`
* [x] The section at the end still has "Figure [2.2](http://epidist.epinowcast.org/articles/epidist.html#fig:fitted-lognormal) shows…" and an empty figure caption that need to be filled out
* [ ] Attend to all comments of @parksw3 in https://github.com/epinowcast/epidist/issues/57#issuecomment-2128054115
  * [x] Define what we mean by primary vs secondary event
  * [ ] Add more citations: I don't think we need to do this
  * [x] Explain Figure 1.4 a bit more carefully
  * [ ] Figure 2.1 change to density: unsure we should do this

## Checklist

### DIFF
--- a/vignettes/epidist.Rmd
+++ b/vignettes/epidist.Rmd
@@ -201,9 +201,10 @@ With our censored, truncated, and sampled data, we are now ready to try to recov
 
 # Fit the model and compare estimates {#fit}
 
-With only access to `obs_cens_trunc_samp`, the delay distribution we observe is biased (Figure \@ref(fig:obs-est)) and we must use a statistical model.
+If we had access to the complete and unaltered `obs`, it would be simple to estimate the delay distribution.
+However, with only access to `obs_cens_trunc_samp`, the delay distribution we observe is biased (Figure \@ref(fig:obs-est)) and we must use a statistical model.
 
-(ref:obs-est) The histogram of delays from the complete retrospective data (with censoring) match closely the underlying distribution, whereas those from `obs_cens_trunc_samp` are systematically biased.
+(ref:obs-est) The histogram of delays from the complete retrospective data `obs_cens` match closely the underlying distribution, whereas those from `obs_cens_trunc_samp` are systematically biased. In this instance, while the complete, retrospective data are still subject to censoring, the extent of this issue is quite minor as compared with that from right truncation.
 
 ```{r obs-est, fig.cap="(ref:obs-est)"}
 dplyr::bind_rows(
@@ -276,7 +277,7 @@ draws <- extract_lognormal_draws(fit)
 Figure \@ref(fig:fitted-lognormal) shows the lognormal delay distribution obtained using the average of the `meanlog` and `sdlog` draws.
 Whereas in Figure \@ref(fig:obs-est) the histogram of censored, truncated, sampled data was substantially different to the underlying delay distribution, using `epidist` we have obtained a much closer match to the truth.
 
-(ref:fitted-lognormal) A fitted delay distribution (in blue) as compared with the true delay distribution (in black).
+(ref:fitted-lognormal) A fitted delay distribution (in pink) as compared with the true delay distribution (in black).
 
 ```{r fitted-lognormal, fig.cap="(ref:fitted-lognormal)"}
 ggplot() +
--- a/vignettes/epidist.Rmd
+++ b/vignettes/epidist.Rmd
@@ -199,7 +199,7 @@ With our censored, truncated, and sampled data, we are now ready to try to recov
 
 # Fit the model and compare estimates {#fit}
 
-If we had access to the data `obs`, then it would be simple to estimate the delay distribution.
+If we had access to the full data `obs`, then it would be simple to estimate the delay distribution.
 However, with only access to `obs_cens_trunc_samp`, the delay distribution we observe is biased (Figure \@ref(fig:obs-est)) and we must use a statistical model.
 
 (ref:obs-est) The histogram of delays from `obs` matches closely the underlying distribution, whereas those from `obs_cens_trunc_samp` are systematically biased.
@@ -208,15 +208,19 @@ However, with only access to `obs_cens_trunc_samp`, the delay distribution we ob
 dplyr::bind_rows(
   obs |>
     dplyr::mutate(type = "Full data") |>
-    dplyr::select(delay, type),
-  obs_cens_trunc |>
+    dplyr::select(delay, type) |>
+    dplyr::mutate(delay = floor(delay)),
+  obs_cens_trunc_samp |>
     dplyr::mutate(type = "Censored, truncated,\nsampled data") |>
-    dplyr::select(delay, type)
+    dplyr::select(delay = delay_daily, type)
 ) |>
+  dplyr::group_by(type, delay, .drop = FALSE) |>
+  dplyr::summarise(n = dplyr::n()) |>
+  dplyr::mutate(p = n / sum(n)) |>
   ggplot() +
-  geom_histogram(
-    aes(x = delay, y = ..density.., fill = type, group = type),
-    position = "dodge"
+  geom_col(
+    aes(x = delay, y = p, fill = type, group = type),
+    position = position_dodge2(preserve = "single")
   ) +
   scale_fill_manual(values = c("#0072B2", "#D55E00")) +
   geom_function(
--- a/vignettes/epidist.Rmd
+++ b/vignettes/epidist.Rmd
@@ -154,7 +154,7 @@ Here we suppose that the interval is daily, meaning that only the date of the pr
 obs_cens <- obs |> observe_process()
 ```
 
-(ref:cens) Interval censoring of the primary and secondary event times obscures the delay times. While daily censoring is most common, `epidist` supports the primary and secondary events having other delay intervals.
+(ref:cens) Interval censoring of the primary and secondary event times obscures the delay times. Many exact delay times (which can take any value) have the same censored delay times (which are restricted to a number of days). While daily censoring is most common, `epidist` supports the primary and secondary events having other delay intervals.
 
 ```{r cens, fig.cap="(ref:cens)"}
 ggplot(obs_cens, aes(x = delay, y = delay_daily)) +
--- a/vignettes/epidist.Rmd
+++ b/vignettes/epidist.Rmd
@@ -204,7 +204,7 @@ With our censored, truncated, and sampled data, we are now ready to try to recov
 If we had access to the complete and unaltered `obs`, it would be simple to estimate the delay distribution.
 However, with only access to `obs_cens_trunc_samp`, the delay distribution we observe is biased (Figure \@ref(fig:obs-est)) and we must use a statistical model.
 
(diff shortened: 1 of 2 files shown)


athowes:
> Figure 2.1: Just so I know I'm understanding this properly, do the orange bars represent full data without censoring? I wonder if it makes more sense to plot them with density plots (alongside the true density) since they're continuous data? We could also consider plotting something like the mean (vertical line) to show the bias.

> Agree - I think we don't then plot correctly discretised dists which we probably should.

Regarding this point, even if the data can be any value (continuous) I think a histogram can still be preferred to a density as a density relies on some KDE parameters (and so ultimately is a model of the data rather than a "true" representation, while a histogram is more of a "true" representation).

For this reason not clear to me that it's better to move to density. I do perhaps see that you'd want the histogram bins to line up with "daily" if that's what you mean also?

parksw3:
Actually, I just took a quick look at it (thought there would be a lot more to review so I suggested Thursday, sorry).

- For simulating exponential cases, I wonder if it makes sense to say that r is the growth rate, rather than the rate parameter. We're not quite simulating an exponential distribution because it's bounded.
- Not sure how to link code but L139: `delay <- ptime <- stime <- NULL` I'm not sure why this line is necessary in `simulate_secondary`?

athowes:
I found a bug in my code (I used `obs_cens_trunc` rather than `obs_cens_trunc_samp`) and have updated the barplot to be at a daily level (using `geom_col`):

image

I have not yet updated away from using the "full data" (towards using the "censored data" as suggested by Sam / above). Here the orange bars are showing a different bias as to before in the rendered version.

athowes:
Here's my light investigation into the discussion about histograms being biased. To me, the final one with highest sample size looks pretty good / unbiased.

image

```r
set.seed(123)

plot_hist <- function(n) {
  data.frame(value = rnorm(n, mean = 0, sd = 1)) |>
    ggplot(aes(x = value)) +
    geom_histogram(aes(y = ..density..), bins = 30, color = "black", fill = "grey90") +
    stat_function(fun = dnorm, args = list(mean = 0, sd = 1), color = "firebrick", size = 1) +
    labs(title = paste0("Sample size: ", n), y = "", x = "") +
    theme_minimal()
}

plots <- lapply(10^{0:5}, plot_hist)
patchwork::wrap_plots(plots)

```

seabbs (commented):
I suggest moving to plotting the retrospective observed data here and opening an issue for exploring how to plot the underlying continuous data in a way that everyone agrees is correct.

vignettes/epidist.Rmd
```diff
@@ -210,20 +208,23 @@ However, with only access to `obs_cens_trunc_samp`, the delay distribution we ob
 dplyr::bind_rows(
   obs |>
     dplyr::mutate(type = "Full data") |>
-    dplyr::select(delay, type),
-  obs_cens_trunc |>
+    dplyr::select(delay, type) |>
+    dplyr::mutate(delay = floor(delay)),
```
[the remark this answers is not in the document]
athowes: Yes this made things clearer for me. That this is the floor of the delay, whereas the truncated delay is the difference between two floored numbers.

athowes:
Here is that plot with all the proposed possibilities:

image

athowes:
I've moved this to using the retrospective data.

vignettes/epidist.Rmd
```diff
@@ -201,29 +201,31 @@ With our censored, truncated, and sampled data, we are now ready to try to recov
 
 # Fit the model and compare estimates {#fit}
 
-If we had access to the data `obs`, then it would be simple to estimate the delay distribution.
+If we had access to the complete and unaltered `obs`, it would be simple to estimate the delay distribution.
 However, with only access to `obs_cens_trunc_samp`, the delay distribution we observe is biased (Figure \@ref(fig:obs-est)) and we must use a statistical model.
 
-(ref:obs-est) The histogram of delays from `obs` matches closely the underlying distribution, whereas those from `obs_cens_trunc_samp` are systematically biased.
+(ref:obs-est) The histogram of delays from the complete retrospective data `obs_cens` match closely the underlying distribution, whereas those from `obs_cens_trunc_samp` are systematically biased. In this instance, while the complete, retrospective data are still subject to censoring, the extent of this issue is quite minor as compared with that from right truncation.
```
seabbs, line 207: I don't think we should be downplaying censoring like this IMO. 

The teaching point is that the real-time data doesn't look like the underlying distribution and nor does the fully observed data but usually to a lesser degree. The key point being that both need a model to map to the underlying distribution. 

We can then add another teaching point which is that if we knew the infection time (and stress we likely never will) then observed censored daily data (i.e your daily histogram) we would have no bias (just increased obs error from the censoring and sampling variation).

seabbs: One way to make this point (as I think @parksw3) is suggesting is to plot the mean and sd of the empirically observed data at different times vs the underlying true distribution. you could also add sampling variation to this if you wanted. This would then help you with some of your below points about "being close to the data"

athowes: image

Here's the image that this is a caption for. It looks to me that the main issue here is the trunaction rather than the censoring (here we also have the sampling). So yes, it's a lesser degree, which the text does say. Perhaps I'd prefer to write "relatively" minor rather than "quite" minor.

seabbs: My point here is we are still saying people should use a model in most cases

athowes: > plot the mean and sd of the empirically observed data at different times vs the underlying true distribution. you could also add sampling variation to this if you wanted. 

I think this would add something to understanding. It might be quite a lot of code / complication for the getting started vignette, which I'd prefer to keep simple if possible. Happy to make another issue for this if you'd like.

athowes: I am thinking of adding another sentence along the lines of "that said, we still think it's important to use a model to take into account censoring...". Is there a relevant little section or citation we could point to here for people who are interested? Could also just say in which situations it'd be more of an issue.

athowes: @seabbs if there are any changes you'd like to make to get this over the line I'm happy for you go to into the branch.

seabbs: > I think this would add something to understanding. It might be quite a lot of code / complication for the getting started vignette, which I'd prefer to keep simple if possible. Happy to make another issue for this if you'd like.

Yes that makes sense. I am suggesting this vs trying to pack in more context to the above plot.

> s there a relevant little section or citation we could point to here for people who are interested? Could also just say in which situations it'd be more of an issue.

The long delay paper and Kelly's paper both cover this. 

> "that said, we still think it's important to use a model to take into account censoring.

Yes agree. If we want to make this clearer we can just shift to using a slightly shorter delay as the censoring bias is absolute

athowes: I've rewritten this caption. If there are specific changes you would like me to make please write them down and I will do them.

athowes: Unsure if I should resolve here as the above caption is outdated.

seabbs: Prior to resolving I think it would be useful to either respond to some of the above suggestions (i.e a shorter delay), move to issues, or adapt the current text in response to them

vignettes/epidist.Rmd
```diff
@@ -30,35 +30,36 @@ knitr::opts_chunk$set(
...
-Finally, to run this vignette yourself, you will need the `data.table`, `purrr` and `ggplot2` packages installed.
-Note that to work with outputs from `epidist` you do not need to use `data.table`: any tool of your preference is suitable.
+To run this vignette yourself, as well as the `epidist` package, you will need the `data.table`^[Note that to work with outputs from `epidist` you do not need to use `data.table`: any tool of your preference is suitable!], `purrr`, `ggplot2`, `gt`, and `dplyr` packages installed.
+
+```{r load-requirements}
+library(epidist)
+library(data.table)
+library(purrr)
+library(ggplot2)
+library(gt)
```
seabbs, line 60: We need this as a suggests in the description?

vignettes/epidist.Rmd
```diff
@@ -156,7 +156,7 @@ Here we suppose that the interval is daily, meaning that only the date of the pr
 obs_cens <- obs |> observe_process()
 ```
 
-(ref:cens) Interval censoring of the primary and secondary event times obscures the delay times. While daily censoring is most common, `epidist` supports the primary and secondary events having other delay intervals.
+(ref:cens) Interval censoring of the primary and secondary event times obscures the delay times. Many exact delay times (which can take any value) have the same censored delay times (which are restricted to a number of days). While daily censoring is most common, `epidist` supports the primary and secondary events having other delay intervals.
```
seabbs, line 159: Can you clarify this a bit to me? I'm not sure I follow the intent.

seabbs (approved):
This looks good I think. One question about intent as I am slightly not following and another about package dependencies.

vignettes/epidist.Rmd
seabbs, line 159:
```suggestion
(ref:cens) Interval censoring of the primary and secondary event times obscures the delay times. A common example of this is when events are reported as daily aggregates. While daily censoring is most common, `epidist` supports the primary and secondary events having other delay intervals.
```

I don't think this comment was addressed. I've reworded. If you feel strongly about the previous sentence please clarify why.

vignettes/epidist.Rmd
```diff
@@ -201,29 +201,31 @@ With our censored, truncated, and sampled data, we are now ready to try to recov
 
 # Fit the model and compare estimates {#fit}
 
-If we had access to the data `obs`, then it would be simple to estimate the delay distribution.
+If we had access to the complete and unaltered `obs`, it would be simple to estimate the delay distribution.
 However, with only access to `obs_cens_trunc_samp`, the delay distribution we observe is biased (Figure \@ref(fig:obs-est)) and we must use a statistical model.
```
athowes, line 206: Test

seabbs (approved):
LGTM

PR #70, opened 2024-06-03
S3 structure refactor

### LINKED ISSUE #21: Propose architecture for gamma delays
## Goal
Currently the package only estimates lognormal delays. We'd like to add the ability to estimate gamma delays. We'd like to think through how to architect the ability to estimate more than one kind of model.

Specifically, we need functions that can:

* Estimate gamma and lognormal delays
* Process the output of a gamma or lognormal brms fit and transform/retrun the estimated parameters into {mean/sd; distribution-specific parameters on the natural scale; distribution-specific parameters on the log scale}
* Summarize the transformed outputs into quantiles
* Make some plots of the distribution and parameter estimates

**Describe the solution you'd like**
* The easiest approach is to hard-code separate models and post-processing functions for the gamma and lognormal distributions, but this is clunky and inelegant
* Let's read about R classes and brms families to think through whether there's a better architectural solution that can be implemented with relatively light effort while we're starting

**Scope**
* Read about implementation options
* Draft a proposal for implementation in a new issue
* Tag @seabbs and Katie for review before starting to implement

athowes:
## Description

This is my branch for playing with the S3 refactor. So far:

* Decomposed existing `latent_truncation_censoring_adjusted_delay()` into generics:
  * Data preparation generic `epidist_prepare`
  * Getting supplementary Stan code generic `epidist_stancode`
  * Priors (currently empty) generic `epidist_priors`
  * The formula object generic `epidist_formula`
  * The family object generic `epidist_family`
  * A fitting function generic `epidist`
* Created  methods for the `epidist_ltcad` class (an abbreviation of latent truncation censoring adjusted delay)
  * Calling `obs_prep <- epidist_prepare(obs, model = "ltcad")` then `epidist(obs_prep)` replicates the functionality of `latent_truncation_censoring_adjusted_delay(obs)`
  
### Remaining questions

* Is there a way to get set up with priors for `epidist`? This would be a feature extension i.e. making is easier for the user to provide their own priors. What about just letting them pass in `brms` priors?
* How do the `custom_stancode` and `family` / `formula` generics interact?
  * Different families will need different `custom_stancode`
  * Will different formulas need different Stan code? I don't know. Perhaps not?
* How difficult would it be to do this where the `prepare` generic is called as a part of the `epidist` fitting generic? i.e. in one step. If it's quite hard and limits what we can do, I'm fine to have the two-step process
* I might like to move some of the functions "inside" `epidist::epidist` rather than have everything passed in as arguments. I guess this removes possible functionality from the user but simplifies things? Perhaps it's also going to be in some ways required if the `custom_stancode` is a function of some `family` object...
* Light take: I think we should be approximately getting rid of the `epidist_stancode` method and putting the relevant Stan code and options to modify it in the other relevant methods. If a user really really wants to change the underlying Stan code a lot then they can generate it and modify it themselves rather than interfacing with it via passing long strings.

Closes #21.

## Checklist


athowes:
Call with @seabbs:

* Look at `t[]` notation in Stan (inbuilt truncation)
* There is also `cens` notation, but it only works for single censoring but not double censoring
  * There might be a way around this using "cohort-based" model (this is how they are doing it in `EpiAware`)
* `as.class`, `is.class`, `validate.class`, .. rather than prepare
* Models with overlapping plotting functionality
* Adding in the Stan details for reproducibility is a good idea (and not hard)
* Getting rid of `dry` option if we document well and just provide `fn`
* The main thing the user would want to change is the likelihood -- to discuss later if we can think of any other viable changes
  * Specify the growth rate for all of the primary event priors. Prior influenced by underlying growth rate
* Whether or not we can provide priors for custom parameters -- this might be quite hard and not worth it
* Add issue for naming of models
  * Extra models: exact one, "solving an integral" one, ...

Next steps:

* Getting rid of the Stan bit
* Adding in the likelihood as string

seabbs:
My end of the call: This is looking really good.

NAMESPACE
```diff
@@ -1,5 +1,11 @@
 # Generated by roxygen2: do not edit by hand
 
+S3method(epidist,epidist_ltcad)
```
seabbs, line 3: Issue to finalise name?

NAMESPACE
```diff
@@ -1,5 +1,11 @@
...
+S3method(epidist_family,epidist_ltcad)
+S3method(epidist_formula,epidist_ltcad)
+S3method(epidist_prepare,default)
```
seabbs, line 6: Why is this the only one with a default method? Can we try and get away with a default method for `epidist` that actually does something? For the others can likely be nothing right now or it could just pass through to `brms`

R/ltcad.R
```diff
@@ -0,0 +1,133 @@
+#' @method epidist_prepare epidist_ltcad
+#' @family ltcad
+#' @export
+epidist_prepare.epidist_ltcad <- function(data) {
```
seabbs, line 4: Is the action to move these to as_epdisit_ltcad etc for a new issue? If so we need to make one

R/ltcad.R
```diff
@@ -0,0 +1,133 @@
...
+  id <- obs_t <- obs_at <- NULL
```
seabbs, line 5: I generally prefer to do the NULLing via global variables in utils but I don't feel strongly.

R/ltcad.R
```diff
@@ -0,0 +1,133 @@
+#' @method epidist_prepare epidist_ltcad
```
seabbs, line 1: Need issues to flesh out the docs for these methods

R/ltcad.R
```diff
@@ -0,0 +1,133 @@
...
+#' @param sigma Formula for the delay standard deviation. Defaults to intercept
+#' only.
+#' @method epidist_formula epidist_ltcad
+#' @family ltcad
+#' @export
+epidist_formula.epidist_ltcad <- function(data, delay_central = ~ 1,
+                                          sigma = ~ 1) {
+  delay_equation <- paste0(
+    "delay_central | vreal(obs_t, pwindow_upr, swindow_upr)",
+    paste(delay_central, collapse = " ")
```
seabbs, line 46: I think you can just as.string this for the same effect.

R/ltcad.R
```diff
@@ -0,0 +1,133 @@
...
+  )
+
+  sigma_equation <- paste0("sigma", paste(sigma, collapse = " "))
+  form <- brms::bf(as.formula(delay_equation), as.formula(sigma_equation))
```
seabbs, line 50: we need to do something about our namespace. Are we: 

1. Not using the namespace nad only using `package::function` 
2. Using @importFrom in the roxygen
3. Both all the time
4. 2. and then sometimes 1. for uncommon functions

seabbs: I usually do 4

R/ltcad.R
```diff
@@ -0,0 +1,133 @@
...
+#' @method epidist epidist_ltcad
+#' @family ltcad
+#' @export
+epidist.epidist_ltcad <- function(data, formula = epidist_formula(data),
+                                  family = epidist_family(data),
+                                  priors = epidist_priors(data),
+                                  fn = brms::brm,
+                                  ...) {
+  
+  stanvars_version <- epidist_version_stanvar()
```
seabbs, line 79: Is there any benefit to not at least having the stan be coming from an internal function? My concern here is its a bit long and a bit much to read.

R/ltcad.R
```diff
@@ -0,0 +1,133 @@
...
+  
+  stanvars_functions <- brms::stanvar(
+    block = "functions", scode = epidist_stan_chunk("functions.stan")
+  )
+
+  family_name <- gsub("latent_", "", family$name)
```
seabbs, line 85: these two bits of code look like a function to me

NAMESPACE
athowes: Yes agree. Will make.

R/ltcad.R
```diff
@@ -0,0 +1,133 @@
...
+#' @export
+epidist.epidist_ltcad <- function(data, formula = epidist_formula(data),
+                                  family = epidist_family(data),
+                                  priors = epidist_priors(data),
+                                  fn = brms::brm,
+                                  ...) {
+  
+  stanvars_version <- epidist_version_stanvar()
+  
+  stanvars_functions <- brms::stanvar(
```
seabbs, line 81: Can you summarise what advantages and disadvantages you see having the stan code loading and overwriting within this function vs in its own function and break those down either based on using the old arg input and the other option (an internal function)

R/ltcad.R
```diff
@@ -0,0 +1,133 @@
...
+    ) +
+    brms::stanvar(
+      block = "data",
+      scode = "array[wN] int woverlap;",
+      x = data[woverlap > 0][, row_id],
+      name = "woverlap"
+    )
+  
+  stanvars_parameters <- brms::stanvar(
+    block = "parameters", scode = epidist_stan_chunk("parameters.stan")
```
seabbs, line 111: Noting that the current naming scheme of the stan code doesn't extend to including other models. Again likely a issue.

R/ltcad.R
athowes: We have an existing issue for that. This was to fix the warnings as they appear here. I'll look into the two options on that issue (I don't know how the global variables option works at the moment).

R/ltcad.R
```diff
@@ -0,0 +1,133 @@
...
+  )
+  
+  stanvars_priors <- brms::stanvar(
+    block = "model", scode = epidist_stan_chunk("priors.stan")
+  )
+  
+  stanvars_all <- stanvars_version + stanvars_functions + stanvars_data +
+    stanvars_parameters + stanvars_tparameters + stanvars_priors
+  
+  fit <- fn(
```
seabbs, line 125: If it wasn't for the stan code above this `epidist` function could be a package wide default or even (shock - not s3). Thoughts on that?

R/utils.R
```diff
@@ -0,0 +1,23 @@
+#' Read in a epidist Stan code chunk
+#'
+#' @param path The path within the "stan" folder of the installed epidist
+#' package to the Stan code chunk of interest.
+#' @return A character string containing the Stan code chunk of interest.
+#' @family utils
+#' @export
```
seabbs, line 7: For an issue. We might want this to be internal unless we want to support users loading in the stan code?

R/utils.R
```diff
@@ -0,0 +1,23 @@
...
+#' Label a epidist Stan model with a version indicator
+#'
+#' @return A brms stanvar chunk containing the package version used to build
+#' the Stan code.
+#' @family utils
+#' @export
+epidist_version_stanvar <- function() {
+  version <- utils::packageVersion("epidist")
+  comment <- paste0("// code chunks used from epidist ", version, "\n")
+  stanvar(scode = comment, block = "functions")
```
seabbs, line 22: not using `brms::` or in namespace - likely an issue

inst/stan/functions.stan
```diff
@@ -0,0 +1,7 @@
+real latent_family_lpdf(vector y, vector mu, vector sigma, vector pwindow,
+                           vector swindow, array[] real obs_t) {
+  int n = num_elements(y);
+  vector[n] d = y - pwindow + swindow;
+  vector[n] obs_time = to_vector(obs_t) - pwindow;
+  return family_lpdf(d | mu, sigma) - family_lcdf(obs_time | mu, sigma);
```
seabbs, line 6: Move T[] to issue?

seabbs (commented):
Overall this looks good. 

I would like some justification for the stan code inside epidist and I think we need a bunch of issues to address already discussed things (or discuss them more).

NAMESPACE
seabbs: (by nothing I usually mean a warning saying this class doesn't have x method implemented - this can really help when people are making their own methods)

R/ltcad.R
```diff
@@ -0,0 +1,133 @@
...
+    formula = formula, family = family, stanvars = stanvars_all,
+    backend = "cmdstanr", data = data, ...
+  )
+  
+  class(fit) <- c(class(fit), "epidist_fit")
```
seabbs, line 130: we might also want to add a class based on the family that has been used (for custom postprocessing). -> issue

NAMESPACE
athowes: https://github.com/epinowcast/epidist/issues/72

R/utils.R
seabbs: If we see people making their own models then I see the arg for why it would be internal. However if we do see that it suggests to me the brms wrapping we do should be easier for them to reproduce vs having to copy and paste out of epidist

R/ltcad.R
athowes: Added issue for this https://github.com/epinowcast/epidist/issues/73

R/ltcad.R
athowes: Added issue for this https://github.com/epinowcast/epidist/issues/74

R/ltcad.R
athowes: I copied this from `idbrms`. Do you think `as.string` is better?

R/ltcad.R
athowes: https://github.com/epinowcast/epidist/issues/75

R/ltcad.R
athowes: Added issue https://github.com/epinowcast/epidist/issues/76.

inst/stan/functions.stan
athowes: https://github.com/epinowcast/epidist/issues/77

R/ltcad.R
athowes: Unsure what this is pointing at. Line 85?

R/ltcad.R
athowes: I have moved it to an external function. I put it as external as all the functions are external at the moment and I'm adding an issue about going through and changing things to internal functions.

R/ltcad.R
athowes: I've moved to using its own function. Added issue to discuss this here: https://github.com/epinowcast/epidist/issues/78

R/ltcad.R
athowes: I think we should move it to a package wide default (and have custom versions for models that need them). Opening issue to do this.

R/utils.R
athowes: Adding an issue for which functions to be internal versus external.

R/utils.R
athowes: Added `brms::`

NAMESPACE
athowes: Why -> I copied `idbrms`.

Yes I think we should have default method for `epidist` see #80.

And yes seems reasonable to add default method for non-implemented classes giving a warning. Will add issue.

athowes: Added issue  #82.

R/ltcad.R
athowes: Added issue #81.

athowes:
Thanks @seabbs!

Have:

* Added issues
* Moved to Stan code as function, function as argument to `epidist`. Open to changing this in future #78

R/ltcad.R
athowes: There isn't an `as.string` function (looking quickly) and:

```
> as.character(~1)
[1] "~" "1"
```

i.e. it would still need `paste`.

R/ltcad.R
athowes: This code is a bit ugly because `family$name` is something like `"latent_lognormal"` but what I want to extract out as `family_name` is `"latent"`. That's the string which gets fed in and `gsub`ed next.

R/ltcad.R
seabbs: fair enough. I'll check and see if I've done anything shorter for this before now in a minute.

seabbs: The answer is broadly no. Before I did something a bit more convoluted: https://github.com/epinowcast/epinowcast/blob/886b45cb4bc5f338fa53d22e83d25335c55b1a4a/R/formula-tools.R#L90

R/ltcad.R
seabbs: I'm pointing at the code used to manipulate family name

seabbs (approved):
Nice looks good. This seems like a really solid basis to be getting started with. Happy all comments are now issues.

Issue #84, opened 2024-06-06
Move `inst/s3-testing.R` somewhere


athowes:
Probably shouldn't have random testing files in `inst/`. Could:

1. Delete
2. Move to a test
3. Move somewhere else

seabbs:
Move to test I think

athowes:
Agree on moving to test. Will need to define what test is appropriate. I've just been using it as "Stan runs fine" so I imagine there are easy tests to add on that.

Have also added other small test snippets in there which should be moved to formal tests.

athowes:
Closed via #106.

PR #52, opened 2024-05-21
Issue 48: Ebola vignette

athowes:
## Description

This PR will close #48. The point of this vignette is to demonstrate some of the more advanced capabilities of the `epidist` package.

## Checklist

- [ ] My PR is based on a package issue and I have explicitly linked it.
- [ ] I have included the target issue or issues in the PR title in the for Issue(s) *issue-numbers*: PR title
- [ ] I have read the [contribution guidelines](https://github.com/epinowcast/.github/blob/main/CONTRIBUTING.md).
- [ ] I have tested my changes locally.
- [ ] I have added or updated unit tests where necessary.
- [ ] I have updated the documentation if required.
- [ ] My code follows the established coding standards.
- [ ] I have added a news item linked to this PR.
- [ ] I have reviewed CI checks for this PR and addressed them as far as I am able.

Issue #23, opened 2024-05-13
Write an inference test


kgostic:
## We want to test that the models are estimating the correct parameter values

* Generate a synthetic dataset
* Set up an inference test
* Set it up to run in CI on pull request?

## Scope
* Should this be an end-to-end test that includes pre- and post- processing or just an inference test?

athowes:
Closed via #113.

PR #95, opened 2024-06-09
Issue 71: Add help for passing priors

### LINKED ISSUE #71: Facilitate user passing in priors
## Goal

As a part of a Bayesian analysis, users of this package should be able to easily pass in prior distributions they would like to use. We should also provide sensible default options.

## Context

`epidist` interfaces via `brms` making it easy to help users pass in some priors. For other parameters (custom) then it's harder.

## Required features

- Update `epidist_priors` to document the names of the parameters that users can pass in
- Update `epidist::epidist` to accept input of priors
- Possibly change the default `brms` priors if we think there is good reason to   

## Out of scope

- Likely out of scope to allow for passing in priors for custom parameters (not the `brms` default ones)

## Related documents

- https://github.com/epinowcast/epidist/pull/70

athowes:
## Description

This PR closes #71.

## Checklist


R/ltcad.R
```diff
@@ -68,6 +61,55 @@ epidist_family.epidist_ltcad <- function(data, family = "lognormal") {
...
+  }
+
+  message(
+    "Warning: No suitable parameters found for the user provided prior(s):\n",
+    paste(replace_fail_message, collapse = "\n")
+  )
+  
+  return(brms:::as.brmsprior(rbindlist(priors_list)))
+}
+
```
athowes, line 106: @seabbs what do you think of this approach?

athowes: Sorry this comment should refer to the whole function not just this end part.

R/ltcad.R
```diff
@@ -68,6 +61,55 @@ epidist_family.epidist_ltcad <- function(data, family = "lognormal") {
   )
 }
 
+#' Define priors for the LTCAD model
+#'
+#' @method epidist_priors epidist_ltcad
+#' @family ltcad
+#' @export
+epidist_priors.epidist_ltcad <- function(priors_list, data,
+                                         family = epidist_family(data),
```
athowes, line 72: Possibly we want to change things so that the user can pass in priors like `function(prior1, prior2, prior3, data, family)` then we put them into a list

athowes:
Does `brms` give you a current warning if you pass in an object which is not a prior?

athowes:
Sam writes:

>  It looks like the prefit drop in would be https://www.rdocumentation.org/packages/brms/versions/2.19.0/topics/get_prior. so the flow would be run epidist with this to see the current priors and then run it again with new priors supplied to prior (or provide a helper to merge with any custom defaults we run) and then fit. in terms of name linking etc I think it should be addressed by documentation as that will leave the system basically the same as that in brms which is nicely documented

athowes:
* I have documented the `epidist_priors` function, but there aren't tests. It is just a function which returns `NULL`
* I agree with having unit tests starting, but I think place to do that is a PR for #84, #23.
* Perhaps I could add a message in `epidist_priors` saying "using default priors"?

athowes:
Once #98 then something like:

```
class(family) <- "lognormal"
class(data) <- "latent_individual"

epidist_priors(data, family) {
priors <- priors(data)
family <- priors(family)
}
```

seabbs:
> I agree with having unit tests starting, b

We need a specific issue then for the unit tests that have been missed in this PR (unit tests != integration tests)

seabbs:
> Perhaps I could add a message in epidist_priors saying "using default priors"?

Suggest "informative" messaging is its own issue. We use `cli` in `{epinowcast}` and I like it.

seabbs:
> class(family) <- "lognormal"
class(data) <- "latent_individual"

epidist_priors(data, family) {
  priors <- priors(data)
  family <- priors(family)
}

OR if we are saying family captures everything about itself we have priors as part of `epidist_family` along with any custom family function, stan code etc.

athowes:
I'd consider anything related to adding family classes to be its own issue and out of scope for this PR (https://github.com/epinowcast/epidist/issues/98).

So for now, IMO the remaining feature to close this PR is to:

1. Add in better default priors for the mean and SD intercept parameters. (Run some kind of prior predictive to get something reasonable here.)

Remaining point of uncertainty about whether the user should be able to pass anything into the `epidist_priors` function to edit these priors (e.g. numbers for the mean and SD say if it's a Gaussian prior). Recall that the user always has the option to fit and then extract priors and refit (needs to be documented in a vignette [it's already in there in the code documentation but vignette useful too] -- I think the Ebola one #52).

athowes:
```

fit <- epidist(
  data = prep_obs,
  formula = formula,
  family = family,
  priors = priors,
  stancode = stancode,
  sample_prior = "only"
)

draws <- epidist::extract_lognormal_draws(fit)
plot(draws$meanlog)
```

image

I'm guessing `meanlog` isn't a million miles from being `mean = exp(meanlog)`. Though not exactly:

```
nat_dt <- nat_dt[,mean := exp(meanlog + sdlog ^ 2 / 2)]
```

Anyway, the point is that `exp(40)` is not acceptable.

seabbs:
>  Recall that the user always has the option to fit and then extract priors and refit (

I think the workflow we want to encourage is extracting priors prior to fitting which I think `brms` also supports

seabbs:
This all looks reasonable to me. Do you have a proposal for a weakly informed default here? Also what is happening with this being a draft vs ready for review? Usually I try and keep the majority of discussion of the plan to the issue FYI but I don't know if good practice here is set in stone.

seabbs:
My takeway here is maybe we don't need the heavy tails from the student t in the default prior and should up to a normal on the rational that its on the log scale anyway and so will end up right skewed regardless?

seabbs:
Here is some useful info on the default priors in `brms`: https://discourse.mc-stan.org/t/default-student-t-priors-in-brms/17197

athowes:
Yes sorry this PR has been a bit of a mess in terms of comments. Happy to chat about this F2F to figure out better plan for next time.

Yes I agree and have been testing Gaussian priors with much smaller mean and variance and it seems to be better.

athowes:
> I think the workflow we want to encourage is extracting priors prior to fitting which I think `brms` also supports

Yep:

```
> default_prior <- epidist::epidist(data = prep_obs, fn = brms::get_prior)
> default_prior
                prior     class coef group resp  dpar nlpar lb ub  source
   student_t(3, 5, 3) Intercept                                   default
 student_t(3, 0, 2.5) Intercept                 sigma             default
 ```
 
 Though perhaps I'd be expecting it to extract out the new prior I'm putting in (maybe bug and I've not got it right yet).

**Comment by athowes** [comment]
I think ready for review. Remaining points:

* All tests should pass (going to lunch, will update if/when they do)
* Should we add in functionality for user to change default priors in `epidist_prior`?
* I think that `brms::get_prior` does not take as input `prior` and so will just give you the `brms` default priors rather than those that we set. Is there some way to get around this? So basically I don't think the "extract priors without fitting" workflow is working yet https://github.com/epinowcast/epidist/pull/95#issuecomment-2162753853

**Comment by seabbs** [comment]
> 'd consider anything related to adding family classes to be its own issue and out of scope for this PR (https://github.com/epinowcast/epidist/issues/98).

This is fine but it needs to be either detailed in that issue or its linked issue.

> Though perhaps I'd be expecting it to extract out the new prior I'm putting in (maybe bug and I've not got it right yet).

This sounds like a bug. 

> So basically I don't think the "extract priors without fitting" workflow is working yet 

If it doesn't this seems like a bug with `brms` have you checked the priors you are passing are otherwise showing up in the model as you would expect?

File: NAMESPACE [hunk 3]
```diff
@@ -23,7 +23,7 @@ export(epidist)
 export(epidist_family)
 export(epidist_formula)
 export(epidist_prepare)
-export(epidist_priors)
+export(epidist_prior)
```
seabbs, line 26: why the name change?

R/latent_individual.R
```diff
@@ -70,6 +64,31 @@ epidist_family.epidist_latent_individual <- function(data, family = "lognormal",
...
+#' resulting object passed to `epidist` to be used. An example of this workflow
+#' will be provided as a part of a vignette at a future date.
+#'
+#' @inheritParams epidist_prior
+#' @method epidist_prior epidist_latent_individual
+#' @family latent_individual
+#' @export
+epidist_prior.epidist_latent_individual <- function(data, ...) {
+  prior1 <- brms::prior("normal(2, 0.5)", class = "Intercept")
+  prior2 <- brms::prior("normal(0.5, 0.5)", class = "Intercept", dpar = "sigma")
```
seabbs, line 88: if this is one the log scale is it a reasonable prior?

NAMESPACE
athowes: Because `brms` uses `prior` as its argument rather than `priors` (and so I didn't like `prior = priors`)

R/latent_individual.R
athowes: I'm not sure what you mean here?

athowes: Do you have an idea about what mean on the real scale you'd like? I just picked some reasonable seeming numbers but I'm open to changing (I didn't really use my expertise about delays because I don't have much). Of course we should be having something pretty wide.

A more sophisticated version I'd like what 5% and 95% numbers then back calculate to get the two priors.

seabbs: my point is that I don't think this is a prior on sdlog  but on sdlog_log.

NAMESPACE
seabbs: cool that is reasonable. Did you flag this above and I missed it? If not this kind of stuff is super handy to clearly note even if it seems obvious

athowes: No I missed flagging. Noted it's good to flag this type of thing

athowes:
Based on call with Katie:

* Based on looking at `?brms::get_prior` don't think `brms::get_prior` is going to work. Seems to be connected only to defaults
* I think that the best solution is something which uses `brms::get_prior` inside the `epidist_prior` function and then replace the parts of it that we would like to overwrite. We don't let the user pass to `epidist_prior` but they can overwrite the parts of it manually outside if they like

```
epidist_prior <- function() {
  defaults <- brms::get_prior()
  # check here that defaults actually contains the two intercepts?
  defaults[1] <- our_new_normal_prior
  defaults[2] <- our_new_normal_prior2
}

my_prior <- epidist_prior()
defaults[3] <- whatever_user_wants
defaults[1] <- whatever_user_wants_if_they_disagree_with_us
```

What do you think @seabbs and @kgostic?

R/latent_individual.R
athowes: Yes that sounds right to me and I'll change things to fit with it being on `sdlog_log`

seabbs:
Have you looked into the `brms` documentation to see if there is an option to retrieve the non-default priors and if so and not found have you opened a features issue with brms? I think that is the logical step here. 

> I think that the best solution is something which uses brms::get_prior inside the epidist_prior function and then replace the parts of it that we would like to overwrite. We don't let the user pass to epidist_prior but they can overwrite the parts of it manually outside if they like:

I don't understand the logic for this? Why can't users use the standard brms approach to writing priors?

seabbs:
I really think we are overthinking this for now and can shunt complexity to an issue.

athowes:
> Have you looked into the brms documentation to see if there is an option to retrieve the non-default priors and if so and not found have you opened a features issue with brms? I think that is the logical step here.

Yes good point, there is a function which does that:

```
> brms::prior_summary(fit)
            prior     class coef group resp  dpar nlpar lb ub source
   normal(2, 0.5) Intercept                                     user
 normal(0.5, 0.5) Intercept                 sigma               user
 ```
 
 > I don't understand the logic for this? Why can't users use the standard brms approach to writing priors?

I don't understand the sense in which this is not using the standard `brms` approach to writing priors. Under the proposed function, `epidist_prior` would return our default priors for their model in standard `brms` format. The user may interact with that object in any way they want, or they can ignore it and build their own prior (using standard `brms` prior construction tools) if they prefer.

The the thing that I am proposing by moving from `epidist_priorA` to `epidist_priorB` is that the `epidist_priorB` will display all of the priors for the model (including those which we do not know about and will be inheriting the `brms` defaults). This seems better to me as it would allow the workflow to call `epidist_prior` then modify it (using the approach to modifying default priors suggested in `brms` documentation).

```
epidist_priorA <- function() {
  prior1 <- brms::prior("normal(2, 0.5)", class = "Intercept")
  prior2 <- brms::prior("normal(0.5, 0.5)", class = "Intercept", dpar = "sigma")
  return(prior1 + prior2)
}
```

```
epidist_priorB <- function() {
  defaults <- brms::get_prior()
  # check here that defaults actually contains the two intercepts?
  defaults[1] <- "normal(2, 0.5)"
  defaults[2] <- "normal(0.5, 0.5)"
}
```

> If the user doesn't want our defaults or the defaults in brms then they can just pass a new brms prior object to prior?
 
 Yes correct.
 
 Happy to call to clear this up if easier.
 
 Also, don't worry about the linting here, there will be a PR today to fix it all (#102).

**Comment by athowes** [comment]
> I don't understand the logic for this? Why can't users use the standard brms approach to writing priors? If the user doesn't want our defaults or the defaults in brms then they can just pass a new brms prior object to prior?

I expect "knowing exactly which parameters are in the model" to be a barrier to specifying priors for users. Even for me, after working on this package for a month, I wouldn't have been able to immediately write

```
prior1 <- brms::prior("normal(2, 0.5)", class = "Intercept")
prior2 <- brms::prior("normal(0, 0.5)", class = "Intercept", dpar = "sigma")
```

without first seeing what the internal default prior looks like. (What's `dpar`? Why is there no `dpar` for the mean?)

My takeaway from this is that our prior specification features should help the user to input valid priors. One way I can think of to do this is for `epidist_prior` to return all default priors in the model (this would be equivalent to the `epidist_priorB` version I wrote above. Users could then either write their own priors using `brms::prior` (after they know which parameters to set for) or could edit the `prior <- epidist_prior()` object if it's easier.

File: R/latent_individual.R [hunk 5]
```diff
@@ -71,6 +64,35 @@ epidist_family.epidist_latent_individual <- function(data, family = "lognormal",
...
+#' very large delays on the real scale. In particular:
+#'
+#' * `eta_meanlog ~ normal(2, 0.5)`
+#' * `eta_sdlog ~ normal(0, 0.5)`
+#' 
+#' Note that the link function used for `meanlog` is a constant, and the link
+#' function used for `sdlog` is a logarithm, such that the expectation of
+#' `sdlog` is the exponential of `eta_sdlog`.
+#' 
+#' To alter the priors for this model, we recommend first fitting using these
```
seabbs, line 80: I don't think we recommend first fitting do we? We are currently recommending using `get_prior` to `fn` to get the default priors?

seabbs (commented):
This looks good to me aside from the linting and the slightly incorrect suggest prior workflow.

seabbs:
I like the current workflow A as it remains closest to the normal `brms` workflow and is lightweight. I suggest we open an issue with `brms` about if `get_prior` can be modified to include all priors as this seems like it would resolve most reasons to want workflow B. 

> without first seeing what the internal default prior looks like. (What's dpar? Why is there no dpar for the mean?)

My push back to this is that this is the main documented way suggest to change `brms` priors. We may or may not like it but a low effort solution on our end is to stick as closely to this as possible and link out to the documentation (dpar is distribution parameter I believe which is fairly clearly documented in `brms`). 

> I expect "knowing exactly which parameters are in the model" to be a barrier to specifying priors for users. Even for me, after working on this package for a month, I wouldn't have been able to immediately write

After you knew about `get_prior` and the workflow that passes it in via `epdist` function would you have been able to do this?

R/latent_individual.R
athowes: I don't think that `get_prior` as `fn` works. `get_prior` has the wrong inputs and will just return the default `brms` prior. What is better is `prior_summary` but that takes the fitted model as input. So there is something that needs to be changed here (suggest `prior_summary` not `get_prior`). At the moment I don't think there is any function which works as a drop in to `fn` that would get the prior without first fitting (perhaps this is something we can raise a `brms` issue for or somehow write our own wrapper to do it).

athowes: Fixed `prior_summary` change in [f017a16](https://github.com/epinowcast/epidist/pull/95/commits/f017a160fa83e8eb44a26ba07feeef2dac388581).

seabbs: I don't think I do recommend this? I think we should suggest looking at the defaults first without fitting + any we specify and then if needing a more detailed view fitting and looking at prior_summary?

athowes: > 1. Suggest looking at the defaults first without fitting + any we specify

> 2. If needing a more detailed view fitting and looking at `prior_summary`

Note that by including the defaults into `epidist_prior` I think would allow people not to have to do 2. and they could just do 1. (calling `epidist_prior`) to get all the prior information they would need to know to be able to edit the model.

(Basically having to call `get_prior`, see some priors on things, then also call `epidist_prior` and know that the `epidist_prior` ones superseed the `get_prior` ones is a confusing workflow which would be simplified by what I propsed in `epidist_priorB` above.)

athowes: For the time being I'm happy to write out both of these options:

You can either look at the default priors, and look at our replacement priors (without fitting the model).

OR

If you have fitted the model, you can extract all the priors (with our defaults superseeding).

athowes: I've updated the `roxygen2` to list both possible workflows.

athowes:
Added issue for my suggestion about the alternative `epidist_prior` (#108) and listed both workflows in the `roxygen2`. Will monitor checks but I think should be ready to merge!

seabbs (approved):

PR #152, opened 2024-07-12
Issue 151: data validation as S3 generic

### LINKED ISSUE #151: `validate` as S3 method
> We might want an issue to make this an s3 method - that way epdist.default can call a validation function on its input to check its a implemented model

_Originally posted by @seabbs in https://github.com/epinowcast/epidist/pull/125#discussion_r1674115715_

My understanding is that if we move to `validate` as a generic and `validate.epidist_latent_individual` as a method then we can put `validate` into `epidist::epidist`. This seems good to me.

athowes:
## Description

This PR closes #151.

I named it `epidist_validate` following convention of other generics in the package.

## Checklist


R/defaults.R
```diff
@@ -1,3 +1,16 @@
+#' Default method for data validation
+#'
+#' @inheritParams epidist_validate
+#' @param ... Additional arguments for method.
+#' @family defaults
+#' @export
+epidist_validate.default <- function(data, ...) {
+  stop(
+    "No epidist_formula method implemented for the class ", class(data), "\n",
```
seabbs, line 9: we might want to say no validate method defined vs formula as there are more than just formula methods you would need. At a later date (i.e. this is another issue) we should add a link how to some kind of documentaiton on how to make a model class

seabbs (approved):
LGTM

R/defaults.R
athowes: This is just a forget to edit not an intentional.

Issue #131, opened 2024-06-25
Think more cafefully about custom families


athowes:
Currently `epidist_family` for the `epidist_latent_individual` is defined as follows

```
#' @method epidist_family epidist_latent_individual
#' @family latent_individual
#' @export
epidist_family.epidist_latent_individual <- function(data, family = "lognormal",
                                                     ...) {
  brms::custom_family(
    paste0("latent_", family),
    dpars = c("mu", "sigma"),
    links = c("identity", "log"),
    lb = c(NA, 0),
    ub = c(NA, NA),
    type = "real",
    vars = c("pwindow", "swindow", "vreal1"),
    loop = FALSE
  )
}
```

In working on #114 I've been trying to extract out the posteriors for the gamma distribution so that they match the input simulation settings. I've not found this immediately easy to do and I am suspecting there may be complications we are not currently taking into account. By not find this easy I mean I've not been able to find a transformation of the samples which gets the simulated settings back. This could either be that the model isn't getting the right answer or that there is some issue with link functions etc.

I think that both the parameters of the gamma distribution have domain $\mathbb{R}^+$ i.e. non-negative In this code we hard code in that the link functions are `links = c("identity", "log")`. This seems questionable (to have identity link for something non-negative. I suspect this is causing problems with the sampler. As a solution we might want to enable users to have more control of the link functions.

While I'm unsure about this, I'm going to remove the gamma parameter recovery integration test from #114:

```
test_that("epidist.epidist_latent_individual recovers the simulation settings for the delay distribution in the gamma delay case", { # nolint: line_length_linter.
  fit_gamma <- epidist(
    data = prep_obs_gamma,
    family = epidist_family(prep_obs_gamma, family = "gamma"),
  )
  draws <- posterior::as_draws_df(fit_gamma$fit)
  # draws$Intercept
  # draws$Intercept_sigma
})
```

As a part of a solution to this issue, I think we should add it back (passing).

athowes:
I think here I also need to understand the parameterisation of the gamma distribution that `brms` uses.

athowes:
Furthermore, this test fails with 10% divergent transitions:

```
test_that("epidist.epidist_latent_individual fits and the MCMC converges for a gamma delay distribution", { # nolint: line_length_linter.
  fit_gamma <- epidist(
    data = prep_obs_gamma,
    family = epidist_family(prep_obs_gamma, family = "gamma"),
  )
  expect_s3_class(fit_gamma, "brmsfit")
[...]

PR #155, opened 2024-07-12
Issue 99: Informative messaging for priors

### LINKED ISSUE #99: Add informative messaging (for priors minimum)
> Perhaps I could add a message in epidist_priors saying "using default priors"?

Suggest "informative" messaging is its own issue. We use `cli` in `{epinowcast}` and I like it.

_Originally posted by @seabbs in https://github.com/epinowcast/epidist/issues/95#issuecomment-2160160720_

Also should think about other places in the package we would like more messaging. Could start looking through `epinowcast` and copy places messaging is provided there.

athowes:
## Description

This PR closes #99.

## Checklist

### DIFF
--- a/R/latent_individual.R
+++ b/R/latent_individual.R
@@ -213,12 +213,15 @@ epidist_prior.epidist_latent_individual <- function(data, ...) {
   prior1 <- brms::prior("normal(2, 0.5)", class = "Intercept")
   prior2 <- brms::prior("normal(0, 0.5)", class = "Intercept", dpar = "sigma")
 
-  cli::cli_inform(c(
-    "i" = "The following priors have been set:",
-    "*" = "normal(2, 0.5) on the intercept of the meanlog linear predictor",
-    "*" = "normal(0, 0.5) on the intercept of the sdlog linear predictor",
-    "To alter priors, or set priors on other parameters, see ?epidist_prior."
-  ))
+  cli::cli_inform(
+    message = c(
+      "i" = "The following priors have been set:",
+      "*" = "normal(2, 0.5) on the intercept of distributional parameter mu",
+      "*" = "normal(0, 0.5) on the intercept of distributional parameter sigma",
+      "To alter priors, or set priors on other parameters, see ?epidist_prior."
+    ),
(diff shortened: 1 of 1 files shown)


R/latent_individual.R
```diff
@@ -205,12 +205,21 @@ epidist_family.epidist_latent_individual <- function(data, family = "lognormal",
...
+#' @importFrom cli cli_inform
 #' @export
 epidist_prior.epidist_latent_individual <- function(data, ...) {
   validate_latent_individual(data)
 
   prior1 <- brms::prior("normal(2, 0.5)", class = "Intercept")
   prior2 <- brms::prior("normal(0, 0.5)", class = "Intercept", dpar = "sigma")
+
+  cli::cli_inform(c(
+    "i" = "The following priors have been set:",
```
seabbs, line 217: I think you need to be what the link function is for each or this is confusing.

seabbs (commented):
I'm not sure I would want this level of verbosity but I am happy to go with it if you think its needed. One small issue with needing to define what the link is in order for people to understand what they are looking at.

athowes:
Yes I also think it's a bit verbose. What about an option `verbose = FALSE` and have it only show if set to `TRUE`?

seabbs:
> Yes I also think it's a bit verbose

We should be able to use`cli` so it only shows once per session or once every 8 hours etc. Perhaps that is the way to go?

athowes:
Give message about prior on the transformed more natural scale.

R/latent_individual.R
athowes: I have written something corresponding more directly to what the `brms` prior is. I think that we still have some confusion about the priors and families which we will need to handle (e.g. #131, #171). So I'd be in favour of this PR just saying what there is at the moment, and it's something we can come back to once we sort out how we are handling families.

athowes:
I've added an issue about the additional complexities here: https://github.com/epinowcast/epidist/issues/172

I've added the "display once every 8 hours". It's also possible to do "display on new session" instead if people have a preference.

seabbs (approved):

PR #171, opened 2024-07-16
[DEPRECATED] Issue 131: Family parameters complexity fix

### LINKED ISSUE #131: Think more cafefully about custom families

athowes:
## Description

This draft PR will close #131. This draft PR has been deprecated. I will extract out the relevant parts for new PRs to address #180 and #179.

## Checklist

### DIFF
--- a/R/latent_individual.R
+++ b/R/latent_individual.R
@@ -251,10 +251,23 @@ epidist_stancode.epidist_latent_individual <- function(data,
     scode = epidist_stan_chunk("latent_individual/functions.stan")
   )
 
+  
   family_name <- gsub("latent_", "", family$name)
 
   stanvars_functions[[1]]$scode <- gsub(
     "family", family_name, stanvars_functions[[1]]$scode
+  )
+  
+  stanvars_functions[[1]]$scode <- gsub(
+    "dpars_A",
+    paste(paste0("vector ", family$dpars), collapse = ", "),
+    stanvars_functions[[1]]$scode
+  )
+  
+  stanvars_functions[[1]]$scode <- gsub(
+    "dpars_B",
+    paste(family$dpars, collapse = ", "),
+    stanvars_functions[[1]]$scode
   )
 
   stanvars_data <- brms::stanvar(
--- a/R/latent_individual.R
+++ b/R/latent_individual.R
@@ -152,6 +152,11 @@ epidist_formula.epidist_latent_individual <- function(data, delay_central = ~ 1,
 #' @family latent_individual
 #' @export
 epidist_family.epidist_latent_individual <- function(data, family = "lognormal",
+                                                     dpars = c("mu", "sigma"),
+                                                     links =
+                                                       c("identity", "log"),
+                                                     lb = c(NA, 0),
+                                                     ub = c(NA, NA),
                                                      ...) {
   epidist_validate(data)
   checkmate::assert_string(family)
@@ -168,10 +173,10 @@ epidist_family.epidist_latent_individual <- function(data, family = "lognormal",
 
   brms::custom_family(
     paste0("latent_", family),
-    dpars = c("mu", "sigma"),
-    links = c("identity", "log"),
-    lb = c(NA, 0),
-    ub = c(NA, NA),
+    dpars = dpars,
+    links = links,
+    lb = lb,
+    ub = ub,
     type = "real",
     vars = c("pwindow", "swindow", "vreal1"),
     loop = FALSE
--- a/tests/testthat/setup.R
+++ b/tests/testthat/setup.R
@@ -23,8 +23,7 @@ set.seed(101)
 
 shape <- 2
 rate <- 3
-
-# The mean of the gamma distribution is: shape * rate
+mu <- shape / rate
 
 sim_obs_gamma <- simulate_gillespie() |>
   simulate_secondary(
--- a/tests/testthat/test-int-latent_individual.R
+++ b/tests/testthat/test-int-latent_individual.R
@@ -80,11 +80,30 @@ test_that("epidist.epidist_latent_individual Stan code has no syntax errors and 
 test_that("epidist.epidist_latent_individual fits and the MCMC converges  in the gamma delay case", { # nolint: line_length_linter.
   skip_on_cran()
   set.seed(1)
+  
+  gamma_family <- epidist_family(
+    prep_obs_gamma,
+    family = "gamma",
+    dpars = c("mu", "shape"),
+    links = c("log", "identity"),
+    lb = c(0, 0),
+    ub = c(NA, NA)
+  )
+  
+  gamma_prior <- brms::prior("normal(0, 3)", class = "Intercept")
+  
+  gamma_formula <- epidist_formula(prep_obs_gamma)
+  gamma_formula <- brms::bf(gamma_formula, as.formula("shape ~ 1"))
+  gamma_formula$pforms$sigma <- NULL
+
   fit_gamma <- epidist(
     data = prep_obs_gamma,
-    family = epidist_family(prep_obs_gamma, family = "gamma"),
+    family = gamma_family,
+    prior = gamma_prior,
+    formula = gamma_formula,
     seed = 1
   )
+  
   expect_s3_class(fit_gamma, "brmsfit")
   expect_s3_class(fit_gamma, "epidist_fit")
   expect_convergence(fit_gamma)
--- a/tests/testthat/test-int-latent_individual.R
+++ b/tests/testthat/test-int-latent_individual.R
@@ -89,20 +89,17 @@ test_that("epidist.epidist_latent_individual fits and the MCMC converges  in the
     lb = c(0, 0),
     ub = c(NA, NA)
   )
-  
-  gamma_prior <- brms::prior("normal(0, 3)", class = "Intercept")
-  
-  gamma_formula <- epidist_formula(prep_obs_gamma)
-  gamma_formula <- brms::bf(gamma_formula, as.formula("shape ~ 1"))
-  gamma_formula$pforms$sigma <- NULL
 
+  gamma_formula <- epidist_formula(prep_obs_gamma)
+  gamma_formula$pforms$sigma <- NULL
+  gamma_formula <- brms::bf(gamma_formula, as.formula("shape ~ 0"))
+  
   fit_gamma <- epidist(
     data = prep_obs_gamma,
     family = gamma_family,
-    prior = gamma_prior,
+    prior = brms::prior("normal(0, 3)", class = "Intercept"),
     formula = gamma_formula,
-    seed = 1,
-    fn = brms::make_stancode
+    seed = 1
   )
   
   expect_s3_class(fit_gamma, "brmsfit")
@@ -113,14 +110,39 @@ test_that("epidist.epidist_latent_individual fits and the MCMC converges  in the
 test_that("epidist.epidist_latent_individual recovers the simulation settings for the delay distribution in the gamma delay case", { # nolint: line_length_linter.
   skip_on_cran()
   set.seed(1)
+  
+  gamma_family <- epidist_family(
+    prep_obs_gamma,
+    family = "gamma",
+    dpars = c("mu", "shape"),
+    links = c("log", "log"),
+    lb = c(0, 0),
+    ub = c(NA, NA)
+  )
+  
+  gamma_formula <- epidist_formula(prep_obs_gamma)
+  gamma_formula$pforms$sigma <- NULL
+  gamma_formula <- brms::bf(gamma_formula, as.formula("shape ~ 1"))
+  
+  # Fix the shape intercept to log(2) so that it corresponds to simulated value
+  gamma_prior <- brms::prior("normal(0, 3)", class = "Intercept") +
+    brms::prior("constant(0.6931472)", class = "Intercept", dpar = "shape")
+  
   fit_gamma <- epidist(
     data = prep_obs_gamma,
-    family = epidist_family(prep_obs_gamma, family = "gamma"),
+    family = gamma_family,
+    prior = gamma_prior,
+    formula = gamma_formula,
     seed = 1
   )
+  
+  # What the value is of the delay i.e. mu = 2/3
+  mean(prep_obs_gamma$delay)
+  
   draws_gamma <- posterior::as_draws_df(fit_gamma$fit)
-  # draws_gamma$Intercept
-  # draws_gamma$Intercept_sigma
+  draws_gamma_mu <- exp(draws_gamma$Intercept)
+  
+  expect_equal(mean(draws_gamma_mu), mu, tolerance = 0.1)
 })
 
 test_that("epidist.epidist_latent_individual Stan code has no syntax errors and compiles for an alternative formula", { # nolint: line_length_linter.
--- a/tests/testthat/test-int-latent_individual.R
+++ b/tests/testthat/test-int-latent_individual.R
@@ -85,21 +85,25 @@ test_that("epidist.epidist_latent_individual fits and the MCMC converges  in the
     prep_obs_gamma,
     family = "gamma",
     dpars = c("mu", "shape"),
-    links = c("log", "log"),
+    links = c("log", "identity"),
     lb = c(0, 0),
     ub = c(NA, NA)
   )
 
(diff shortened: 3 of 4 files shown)


athowes:
Example of sampling fail with both `mu` and `shape` parameters being sampled (this is the `mu` intercept):

image

R/latent_individual.R
```diff
@@ -232,11 +237,24 @@ epidist_stancode.epidist_latent_individual <- function(data,
...
+  
   family_name <- gsub("latent_", "", family$name)
 
   stanvars_functions[[1]]$scode <- gsub(
     "family", family_name, stanvars_functions[[1]]$scode
   )
+  
+  stanvars_functions[[1]]$scode <- gsub(
+    "dpars_A",
+    paste(paste0("vector ", family$dpars), collapse = ", "),
```
seabbs, line 249: isn't dpars a vector of length two?

R/latent_individual.R
```diff
@@ -151,6 +151,11 @@ epidist_formula.epidist_latent_individual <- function(data, delay_central = ~ 1,
 #' @family latent_individual
 #' @export
 epidist_family.epidist_latent_individual <- function(data, family = "lognormal",
+                                                     dpars = c("mu", "sigma"),
```
seabbs, line 154: I think mu and sigma are the standard names brms always uses so we might want to keep them?

R/latent_individual.R
seabbs: ah I see so dpars_a doesn't do anything?

R/latent_individual.R
athowes: This is the crux of the issue. What are `mu` and `sigma` for e.g. the Gamma distribution? My understanding is that `mu` is always there,  but I don't know that `sigma` is. For the Gamma I think it's `mu` and `shape` ($\alpha > 0$). More broadly I think it's `mu` plus other things.

tests/testthat/test-int-latent_individual.R
```diff
@@ -68,6 +68,74 @@ test_that("epidist.epidist_latent_individual Stan code has no syntax errors and
...
+  gamma_family <- epidist_family(
+    prep_obs_gamma,
+    family = "gamma",
+    dpars = c("mu", "shape"),
+    links = c("log", "log"),
+    lb = c(0, 0),
+    ub = c(NA, NA)
+  )
+
+  gamma_formula <- epidist_formula(prep_obs_gamma)
```
seabbs, line 84: Its shape and rate right?

R/latent_individual.R
athowes: r.e. better yet just leave as is, if I'm correct about `sigma` not existing for some families then we can't do that. So TBD based on resolution of that discussion.

r.e. why not recursively replace, I'm not sure what exactly that would be. I did `dpars_A` and `dpars_B` because you need `vector dpars` in one place and `dpars, ...` in the other.

tests/testthat/test-int-latent_individual.R
athowes: image

See https://cran.r-project.org/web/packages/brms/vignettes/brms_families.html. I think it's `mu` and `shape`.

seabbs: ah right its mu and shape it seems: https://solomonkurz.netlify.app/blog/2023-05-14-causal-inference-with-gamma-regression-or-the-problem-is-with-the-link-function-not-the-likelihood/

tests/testthat/test-int-latent_individual.R
```diff
@@ -68,6 +68,74 @@ test_that("epidist.epidist_latent_individual Stan code has no syntax errors and
...
+  gamma_formula$pforms$sigma <- NULL
+  gamma_formula <- brms::bf(gamma_formula, as.formula("shape ~ 0"))
+  
+  fit_gamma <- epidist(
+    data = prep_obs_gamma,
+    family = gamma_family,
+    prior = brms::prior("normal(0, 3)", class = "Intercept"),
+    formula = gamma_formula,
```
seabbs, line 92: Having to change the names here makes this not a great experience I think for the user. If we don't want to unpick this suggests moving more things internal to epidist (and passing a character string for family with all current arguments done internally. People would then be able to make code level changes via a new class

tests/testthat/test-int-latent_individual.R
athowes: Yes I'm confused why people are using `identity` link for `mu` when it's got to be positive (?)

tests/testthat/setup.R
```diff
@@ -23,8 +23,7 @@ set.seed(101)
 
 shape <- 2
 rate <- 3
-
-# The mean of the gamma distribution is: shape * rate
+mu <- shape / rate
```
seabbs, line 26: and brms gamma is parameterised as the mean and the shape? not the shape and rate?

tests/testthat/test-int-latent_individual.R
athowes: At the moment I've mostly been trying to get the model to work in the `test-int-latent_individual` script i.e. this code is not a suggestion of the best way to present to the user.

But yes I think these issues do suggest more thought or moving more things internal is required.

tests/testthat/setup.R
athowes: These numbers are fed into `rgamma` in order to simulate data, which takes `shape` and `rate` input

tests/testthat/test-int-latent_individual.R
```diff
@@ -68,6 +68,74 @@ test_that("epidist.epidist_latent_individual Stan code has no syntax errors and
...
+  )
+
+  gamma_formula <- epidist_formula(prep_obs_gamma)
+  gamma_formula$pforms$sigma <- NULL
+  gamma_formula <- brms::bf(gamma_formula, as.formula("shape ~ 0"))
+  
+  fit_gamma <- epidist(
+    data = prep_obs_gamma,
+    family = gamma_family,
+    prior = brms::prior("normal(0, 3)", class = "Intercept"),
```
seabbs, line 91: Have you tried playing around with raw brms and gamma? It looks like it expects the shape to have no link and to use a gamma prior by default? https://rpubs.com/jwesner/gamma_glm

R/latent_individual.R
seabbs: no sorry. As in we keep the names and define them differently. Or we define them as brms does and then need to think about interface in more detail. The thing we don't want to do is try and transform in some mad way!

R/latent_individual.R
seabbs: you can if its just us renaming but I am not saying we should do that. 

I'm not sure I understand why isn't vector just written into the code? Why isn it just injecting `mu, sigma` etc?

seabbs: if you could do this it would let us have standard code for injecting dpars and then not have to make it model specific which I think we want

tests/testthat/setup.R
seabbs: no I'm looking at mu here

tests/testthat/test-int-latent_individual.R
athowes: Yeah interesting. Here is the current Stan code:

```
> fit_gamma
// generated with brms 2.21.6
functions {
  // code chunks used from epidist 0.0.0.9000

  // Here the strings
// * gamma
// * vector mu, vector shape
// * mu, shape
// are/have been replaced using regex

real latent_gamma_lpdf(vector y, vector mu, vector shape, vector pwindow,
                           vector swindow, array[] real obs_t) {
  int n = num_elements(y);
  vector[n] d = y - pwindow + swindow;
  vector[n] obs_time = to_vector(obs_t) - pwindow;
  return gamma_lpdf(d | mu, shape) - gamma_lcdf(obs_time | mu, shape);
}
}
data {
  int<lower=1> N;  // total number of observations
  vector[N] Y;  // response variable
  // data for custom real vectors
  array[N] real vreal1;
  // data for custom real vectors
  array[N] real vreal2;
  // data for custom real vectors
  array[N] real vreal3;
  int prior_only;  // should the likelihood be ignored?
  int wN;
  array[N - wN] int noverlap;
  array[wN] int woverlap;
}
transformed data {
}
parameters {
  real Intercept;  // temporary intercept for centered predictors
  real Intercept_shape;  // temporary intercept for centered predictors
  vector<lower = 0, upper = 1>[N] swindow_raw;
vector<lower = 0, upper = 1>[N] pwindow_raw;
}
transformed parameters {
  real lprior = 0;  // prior contributions to the log posterior
  vector<lower = 0>[N] pwindow;
vector<lower = 0>[N] swindow;
swindow = to_vector(vreal3) .* swindow_raw;
pwindow[noverlap] = to_vector(vreal2[noverlap]) .* pwindow_raw[noverlap];
if (wN) {
  pwindow[woverlap] = swindow[woverlap] .* pwindow_raw[woverlap];
}
  lprior += normal_lpdf(Intercept | 0, 3);
  lprior += student_t_lpdf(Intercept_shape | 3, 0, 2.5);
}
model {
  swindow_raw ~ uniform(0, 1);
pwindow_raw ~ uniform(0, 1);
  // likelihood including constants
  if (!prior_only) {
    // initialize linear predictor term
    vector[N] mu = rep_vector(0.0, N);
    // initialize linear predictor term
    vector[N] shape = rep_vector(0.0, N);
    mu += Intercept;
    shape += Intercept_shape;
    mu = exp(mu);
    shape = exp(shape);
    target += latent_gamma_lpdf(Y | mu, shape, pwindow, swindow, vreal1);
  }
  // priors including constants
  target += lprior;
}
generated quantities {
  // actual population-level intercept
  real b_Intercept = Intercept;
  // actual population-level intercept
  real b_shape_Intercept = Intercept_shape;
}
```

tests/testthat/test-int-latent_individual.R
seabbs: I don't think they are I think they are doing that for shape? See comment about running some raw brms

tests/testthat/test-int-latent_individual.R
seabbs: See comment about starting with a base brms fit

tests/testthat/setup.R
athowes: `mu` is used to compare the chains of `mu` parameters from the NUTS?

athowes:
Here's a version of the model using `alpha` as gamma distributed:

```
> fit_gamma
// generated with brms 2.21.6
functions {
  // code chunks used from epidist 0.0.0.9000

  // Here the strings
// * gamma
// * vector mu, vector shape
// * mu, shape
// are/have been replaced using regex

real latent_gamma_lpdf(vector y, vector mu, vector shape, vector pwindow,
                           vector swindow, array[] real obs_t) {
  int n = num_elements(y);
  vector[n] d = y - pwindow + swindow;
  vector[n] obs_time = to_vector(obs_t) - pwindow;
  return gamma_lpdf(d | mu, shape) - gamma_lcdf(obs_time | mu, shape);
}
}
data {
  int<lower=1> N;  // total number of observations
  vector[N] Y;  // response variable
  // data for custom real vectors
  array[N] real vreal1;
  // data for custom real vectors
  array[N] real vreal2;
  // data for custom real vectors
  array[N] real vreal3;
  int prior_only;  // should the likelihood be ignored?
  int wN;
  array[N - wN] int noverlap;
  array[wN] int woverlap;
}
transformed data {
}
parameters {
  real Intercept;  // temporary intercept for centered predictors
  real<lower=0> shape;  // shape parameter
  vector<lower = 0, upper = 1>[N] swindow_raw;
vector<lower = 0, upper = 1>[N] pwindow_raw;
}
transformed parameters {
  real lprior = 0;  // prior contributions to the log posterior
  vector<lower = 0>[N] pwindow;
vector<lower = 0>[N] swindow;
swindow = to_vector(vreal3) .* swindow_raw;
pwindow[noverlap] = to_vector(vreal2[noverlap]) .* pwindow_raw[noverlap];
if (wN) {
  pwindow[woverlap] = swindow[woverlap] .* pwindow_raw[woverlap];
}
  lprior += normal_lpdf(Intercept | 0, 3);
  lprior += gamma_lpdf(shape | 0.01, 0.01);
}
model {
  swindow_raw ~ uniform(0, 1);
pwindow_raw ~ uniform(0, 1);
  // likelihood including constants
  if (!prior_only) {
    // initialize linear predictor term
    vector[N] mu = rep_vector(0.0, N);
    mu += Intercept;
    mu = exp(mu);
    target += latent_gamma_lpdf(Y | mu, shape, pwindow, swindow, vreal1);
  }
  // priors including constants
  target += lprior;
}
generated quantities {
  // actual population-level intercept
  real b_Intercept = Intercept;
}
```

tests/testthat/test-int-latent_individual.R
athowes: You don't think they are using `indentity` for `shape`? But you also write "It looks like it expects the shape to have no link" -- isn't that the same as identity link (or can be constructed to be the same).

And they are using `identity` for `mu` (sometimes):

image

R/latent_individual.R
athowes: > As in we keep the names and define them differently

I think this leads to solutions like "The thing we don't want to do is try and transform in some mad way!".

> Or we define them as brms does and then need to think about interface in more detail

I feel like this is the direction needed.

R/latent_individual.R
athowes: > I'm not sure I understand why isn't vector just written into the code? 

There could be an arbitrary number of distributional parameters. So what should be written? `vector dpar1, vector dpar2, vector dpar3, ...`.

> if you could do this it would let us have standard code for injecting dpars

I don't understand how this would make it more standard.

>  and then not have to make it model specific which I think we want

I'm not sure the way the solution here isn't standard code for injecting `dpars`. Is there a way in which this is model specific?

tests/testthat/test-int-latent_individual.R
```diff
@@ -77,6 +77,78 @@ test_that("epidist.epidist_latent_individual Stan code has no syntax errors and
...
+
+test_that("epidist.epidist_latent_individual recovers the simulation settings for the delay distribution in the gamma delay case", { # nolint: line_length_linter.
+  skip_on_cran()
+  set.seed(1)
+  
+  gamma_family <- epidist_family(
+    prep_obs_gamma,
+    family = "gamma",
+    dpars = c("mu", "shape"),
+    links = c("log", "log"),
```
athowes, line 122: This should be `identity`

athowes:
* [ ] New issue for restructure
* [ ] New issue for making the gamma work
* [ ] Abandon this PR and move the relevant parts into "make the gamma work" and "restructure"

PR #176, opened 2024-07-17
Issue 161: Note about stochastic test failure

### LINKED ISSUE #161: Possibility of reproducible stochastic test fails
In PR #160 we have included setting the seed in R and the seed passed to `brms` such that repeat runs of the integration tests produce the same answer.

It is possible that the code for the model changes, and repeat runs of the test will always fail (for example if it's a p-value of 0.01 then this will happen in 1% of cases). 

I don't know if there is anything that can be done about this. I'd say that one of these integration tests failing is still a good signal something is up, but if you have really checked then it could just be a random fail.

Could test this with changing the seed and record distribution of test fails.

I've labelled this as "question" but really it's more like "warning" or "something to bear in mind".

athowes:
## Description

This PR closes #161.

I've added a note about stochastic test fails to every name of a stochastic test. I'm open to better solutions than this.

## Checklist

### DIFF
--- a/tests/testthat/test-int-latent_individual.R
+++ b/tests/testthat/test-int-latent_individual.R
@@ -1,8 +1,8 @@
-stochastic_test_message <- paste0(
-  "Note: this test is stochastic. Test failure may be bad luck rather than ",
-  "indicate an issue. As this test is reproducible, you may investigate the ",
-  "distribution of test failures varying the input seed."
-)
+# Note: some tests in this script are stochastic. As such, test failure may be
+# bad luck rather than indicate an issue with the code. However, as these tests
+# are reproducible, the distribution of test failures may be investigated by
+# varying the input seed. Test failure at an unusually high rate does suggest
+# a potential code issue.
 
 prep_obs <- as_latent_individual(sim_obs)
 prep_obs_gamma <- as_latent_individual(sim_obs_gamma)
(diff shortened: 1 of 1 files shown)


tests/testthat/test-int-latent_individual.R
```diff
@@ -1,3 +1,9 @@
+stochastic_test_message <- paste0(
```
seabbs, line 1: I think as this is developer side I would be tempted by just adding a comment to this effect vs printing it into the test message?

athowes: Yeah fair, just a comment at the top of the script?

athowes: Printing it to the test message makes it clear which tests this applies to

athowes: I could also put the comment inside each of the relevant tests

seabbs: > I could also put the comment inside each of the relevant tests

So I think this is what I would do but I don't feel that strongly so your call on what you think is clearer/most useful

athowes: I don't have strong preferences either. I slightly dislike copy pasting the same message, so perhaps put it at the top and then a note to refer to the comment in each function

athowes: I've implemented that approach in c7638e5

seabbs (approved):

PR #201, opened 2024-07-29
Add GitHub action for `cmdstanr`

### LINKED ISSUE #104: Add GitHub action to check model running using `cmdstanr`
We could have a GitHub action to check that our Stan code works with the latest version of `cmdstanr`.

See #65.

This is what the action: https://github.com/epinowcast/epinowcast/blob/main/.github/workflows/check-cmdstan.yaml is doing inside of `epinowcast`.

To make this change we would need to copy over this, change the parts that are `epinowcast` specific. For sure this part:

https://github.com/epinowcast/epinowcast/blob/886b45cb4bc5f338fa53d22e83d25335c55b1a4a/.github/workflows/check-cmdstan.yaml#L60-L77

```
      - name: Compile model and check syntax
        run: |
          model <- epinowcast::enw_model()
          # If the model is not syntactically correct above will fail
          # however it may be correct enougth to compile but still contain
          # soft depreciated syntax and so we check the syntax again below 
          # and test the output.
          message <- capture.output(
            model$check_syntax(pedantic = FALSE),
            type = "message"
          )
          # We can't use TRUE here as pendatic check return lots of false
          # positives related to our use of functions.
          stopifnot(
            length(message) != 0 &&
            all(message == "Stan program is syntactically correct")
          )
        shell: Rscript {0}
```

Do we need to agree on which functions will be tested they can run? What about user input Stan functionality? Basically how many possible variations should we be checking?

Also I don't think we have a function to just compile, so we'd be also running. I imagine this would slow things down. Perhaps there is a way we can use `fn` to just compile and not run?

athowes:
## Description

This draft PR will close #104.

## Checklist

### DIFF
--- a/.github/workflows/check-cmdstan.yaml
+++ b/.github/workflows/check-cmdstan.yaml
@@ -59,7 +59,7 @@ jobs:
 
       - name: Compile model and check syntax
         run: |
-          dummy_obs <- data.table::data.table(ptime = 1, stime = 2,
+          dummy_obs <- data.table::data.table(case = 1L, ptime = 1, stime = 2,
             delay_daily = 1, delay_lwr = 1, delay_upr = 2, ptime_lwr = 1,
             ptime_upr = 2, stime_lwr = 1, stime_upr = 2, obs_at = 100,
             censored = "interval", censored_obs_time = 10, ptime_daily = 1,
@@ -73,7 +73,7 @@ jobs:
             stan_file = cmdstanr::write_stan_file(stancode), compile = FALSE
           )
           message <- capture.output(
-            mod(pedantic = FALSE),
+            mod$check_syntax(pedantic = FALSE),
             type = "message"
           )
           # We can't use TRUE here as pendatic check return lots of false
--- a/.github/workflows/check-cmdstan.yaml
+++ b/.github/workflows/check-cmdstan.yaml
@@ -59,13 +59,21 @@ jobs:
 
       - name: Compile model and check syntax
         run: |
-          model <- epinowcast::enw_model()
-          # If the model is not syntactically correct above will fail
-          # however it may be correct enougth to compile but still contain
-          # soft depreciated syntax and so we check the syntax again below 
-          # and test the output.
+          dummy_obs <- data.table::data.table(ptime = 1, stime = 2,
+            delay_daily = 1, delay_lwr = 1, delay_upr = 2, ptime_lwr = 1,
+            ptime_upr = 2, stime_lwr = 1, stime_upr = 2, obs_at = 100,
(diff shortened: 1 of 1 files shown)


athowes:
@seabbs how do I test whether this is working?

seabbs:
#pushtomain

.github/workflows/check-cmdstan.yaml
```diff
@@ -0,0 +1,90 @@
...
+
+      - name: Compile model and check syntax
+        run: |
+          dummy_obs <- data.table::data.table(ptime = 1, stime = 2,
+            delay_daily = 1, delay_lwr = 1, delay_upr = 2, ptime_lwr = 1,
+            ptime_upr = 2, stime_lwr = 1, stime_upr = 2, obs_at = 100,
+            censored = "interval", censored_obs_time = 10, ptime_daily = 1,
+            stime_daily = 1
+          )
+          dummy_obs <- as_latent_individual(dummy_obs)
```
seabbs, line 68: you need to library call epidist and cmdstan or consistently use package::function

.github/workflows/check-cmdstan.yaml
```diff
@@ -0,0 +1,90 @@
...
+          stancode <- epidist::epidist(
+            data = dummy_obs, fn = brms::make_stancode
+          )
+          mod <- cmdstan_model(
+            stan_file = write_stan_file(stancode), compile = FALSE
+          )
+          message <- capture.output(
+            mod(pedantic = FALSE),
```
seabbs, line 76: is this enough on its own?

.github/workflows/check-cmdstan.yaml
```diff
@@ -0,0 +1,90 @@
...
+          # positives related to our use of functions.
+          stopifnot(
+            length(message) != 0 &&
+            all(message == "Stan program is syntactically correct")
+          )
+        shell: Rscript {0}
+
+      - name: Compile and expose stan functions
+        run: |
+          model <- epinowcast::enw_stan_to_r()
```
seabbs, line 89: do you mean to include this?

seabbs (commented):
I think its quite hard to test so just need a careful sight read. Flagged a few issues.

.github/workflows/check-cmdstan.yaml
athowes: Mistake

.github/workflows/check-cmdstan.yaml
athowes: No

athowes:
I've addressed the flagged issues and checked that at least the code runs in a new script. Would guess we can give it a go now.

seabbs (approved):

Issue #188, opened 2024-07-22
Add functionality for generating individual or strata level delay distribution samples


athowes:
In #170 we added functionality to create `mean` and `sd` columns to `data.table`s which contain parameters from either the lognormal or gamma distributional parameters (with the possibility to implement any other required distributions).

Now we want to add functionality to allow users to easily generate the delay distributions they most care about. A first example of this would be the delay distribution estimate for each individual, or strata, in the model. This object should then naturally be able to be passed into `add_natural_scale_ mean_sd` (and indeed we will likely call it within the added function). I think plotting functionality can also be built around the output.

One question to answer here is whether this should be

1. Individual-level: i.e. as many outputs as there are data points
2. Strata-level: i.e. as many outputs as there are strata, e.g. 1 if intercept only, or say the number of areas if it's a spatial model

In the current unit test:

```
test_that("add_natural_scale_mean_sd.lognormal_samples works with posterior samples from the latent lognormal model", { # nolint: line_length_linter.
  skip_on_cran()
  set.seed(1)
  prep_obs <- as_latent_individual(sim_obs)
  fit <- epidist(data = prep_obs, seed = 1)
  ld_mu <- brms::posterior_linpred(fit, transform = TRUE, dpar = "mu")
  ld_sigma <- brms::posterior_linpred(fit, transform = TRUE, dpar = "sigma") 
```

if found a way to do the individual one. Perhaps we should provide both, but broadly I think the strata is preferred. Also, we should really be enabling people to predit whatever they want if we can. Need to think about how to do this / play with available `brms` functionality.

## Notes

```
# Below contains implementation notes:

# Need to modify this (i.e. "latent_lognormal")
fit$family

# (Which includes these names for the link functions needed)
fit$family$link
fit$family$link_sigma

# To work with these existing brms prediction functions
brms::posterior_epred(fit)
brms::posterior_predict(fit)

pp <- brms::prepare_predictions(fit)
pp$dpars$mu$fe$b
pp$dpars$sigma$fe$b

fit$family$family <- "lognormal"
fit$family$name <- "lognormal"

str(fit$family)
brms::posterior_predict(fit)
```

* Is `brms::prepare_predictions(fit)` used inside `posterior_linpred`?
* Check whether order of classes matter

seabbs:
[...]

PR #207, opened 2024-07-30
Issue 204: Remove family as class

### LINKED ISSUE #204: Use class name for families that avoids clashes
> I think if this is the name taken from brms directly which I think it is it could lead to collisions with other packages. Happy for this to be dealt with in an issue.

_Originally posted by @seabbs in https://github.com/epinowcast/epidist/pull/170#discussion_r1695079568_

e.g. `epidist_lognormal` over `lognormal`.

athowes:
## Description

This PR closes #204.

It turns out that I think actually no where in the code are we using the family class so far, so I removed it.

## Checklist

### DIFF
--- a/R/defaults.R
+++ b/R/defaults.R
@@ -91,7 +91,7 @@ epidist.default <- function(data, formula = list(mu ~ 1, sigma ~ 1),
     stanvars = epidist_stancode, backend = "cmdstanr", data = data, ...
   )
 
-  class(fit) <- c(class(fit), "epidist_fit", family$name)
+  class(fit) <- c(class(fit), "epidist_fit")
 
   return(fit)
 }


seabbs (commented):
How are we going to do add_mean_sd without this or is this a remove for now?

athowes:
`add_mean_sd` doesn't use this anywhere at the moment.

Also, `fit` stores the `family` information in it anyway, so we don't need it (we can just `class(...) <- fit$family$name`).

seabbs:
My point was we need a custom family for samples to then apply the class for add_mean_sd

> Also, fit stores the family information in it anyway, so we don't need it (we can just class(...) <- fit$family$name).

We can however use this for now but it won't work if we want to dispatch on a s3 method

seabbs (approved):

athowes:
It's the samples which should have some kind of class, not the `fit` seems like. Unless it'd be both

seabbs:
it is wether or not you need a family based extraction function to get the samples to have the right class

athowes:
I don't think that you do. Or that the extent to which you do then `brms` can handle it. Will see with implementing the #188

PR #210, opened 2024-07-30
Issue 188: Add delay samples functionality

### LINKED ISSUE #188: Add functionality for generating individual or strata level delay distribution samples

athowes:
## Description

This draft PR will close #188.

## Checklist


athowes:
To do before merge:

* [ ] Add issue about adding functionality for full custom family implementation things
* [ ] Discuss choice of name (maybe different to `delay_samples`). Could use word draws, could mention extracting. Want to be similar to the `summary` function
* [ ] Add issue about summary function built on top of this

athowes:
I think this is ready for review. I have created follow up issues.

Other thing is that for some reason it's failing a Ubuntu `R-CMD-check` but in a way that seems not related to the code changes. Any pointers?

seabbs:
The error is here: 

>  Error ('test-unit-postprocess.R:5:3'): add_mean_sd.lognormal_samples works with posterior samples from the latent lognormal model ──
Error: Error: All list elements must have the same length.

You also have a few unrelated to this PR notes and warnings that we should flag and ideally fix

R/postprocess.R
```diff
@@ -1,3 +1,132 @@
...
+  df[["mu"]] <- as.vector(lp_mu)
+  for (dpar in setdiff(names(pp$dpars), "mu")) {
+    lp_dpar <- brms::get_dpar(pp, dpar = dpar, inv_link = TRUE)
+    df[[dpar]] <- as.vector(lp_dpar)
+  }
+  class(df) <- c(
+    class(df), paste0(sub(".*_", "", fit$family$name), "_samples")
+  )
+  dt <- as.data.table(df)
+  dt <- add_mean_sd(dt)
```
seabbs, line 29: do we want to call this within this function? or have the user do it in post? I am open to an argument for either side

R/postprocess.R
```diff
@@ -1,3 +1,132 @@
+#' Extract samples of the delay distribution parameters
+#'
+#' @param fit A model fit with `epidist::epidist`
+#' @param ... Additional arguments passed to `brms::prepare_predictions`
```
seabbs, line 4: I think the modern way to link is [brms::prepare_predictions()] and I think this will result in working pkgdown links whilst what we have currently won't. Not 100% though. What do you think? If this is the case I guess its a new issue to check and do everywhere.

R/postprocess.R
seabbs: I think postprcessing plotting etc will be easier if we leave it here (as we will know it is part of family_samples) but in terms of a function should do one thing I see the arg that users should call this themselves if they want to

R/postprocess.R
```diff
@@ -1,3 +1,132 @@
...
+#' @family postprocess
+#' @autoglobal
+#' @export
+delay_samples <- function(fit, ...) {
```
seabbs, line 8: As we are living in cmdstanr and posterior world I see the argument for using draws

R/postprocess.R
```diff
@@ -1,3 +1,132 @@
...
+  args <- list(...)
+  if ("newdata" %in% names(args)) {
+    newdata <- brms:::validate_newdata(newdata)
```
seabbs, line 11: I think we might want two user functions rather than overloading what we have? 

One (i.e this one) does extraction of draws based on what was fitted and the other (i.e `predict_delays` etc.) handles having a new data input. Both can produce an object of the same class though and share downstream functionality? What do you think?

seabbs: Alternatively I think we should follow more closely to the standard `predict` or `posterior_predict` naming scheme and always have new data as an argument.

seabbs (commented):
This is looking good. Just naming to discuss I think

R/postprocess.R
athowes: Yes I think this is the way to do it and it should be everywhere. Overall I've been a bit lazy with documentation writing and when we get to 0.1.0 I think there is an overall sweep that should be done to improve

R/postprocess.R
athowes: I'm weakly in favour of having it included here. It could be possible that there are workflows we stop users from doing by not expecting them to use the `add_mean_sd` function much e.g. adding the mean and SD of summaries rather than draws. I also don't think it's much of a downside to have these additional columns

R/postprocess.R
athowes: I think one of the `R CMD CHECK` warnings is that we import `posterior` but don't use it. So unclear how much we are living in `posterior` world

R/postprocess.R
athowes: If a user wants to extract draws of parameters from the model they can use standard `brms` and Stan functionality.

These functions are to extract useful functions of parameters.

I guess to be concrete: what do you mean by "extraction of draws based on what was fitted". Is this the same as `delay_samples` with `newdata = NULL` which will get draws for each data point (individual).

About including `newdata` within the names or not, this was something I was unsure about. I'm happy to include it. It is included in `...` but perhaps it's hard for users to know that. The consideration in my head was like "making it obvious by specially pointing to `newdata`" versus "treating all other arguments of `brms::prepare_predictions` as on an equal footing to `newdata`": who knows maybe users want to use those too? But yes I do think `newdata` can have a special place.

athowes:
Added issue about fixing possible check notes: https://github.com/epinowcast/epidist/issues/217

R/postprocess.R
athowes: I think an argument for draws is that is sounds less like a function. Samples sounds like it's doing something (more than draws somehow).

athowes: I think that's also an argument for the "extract" type word. To make it clear that sampling is not happening here, it's just manipulation of samples.

athowes:
> The error is here:
> 
> > Error ('test-unit-postprocess.R:5:3'): add_mean_sd.lognormal_samples works with posterior samples from the latent lognormal model ──
> > Error: Error: All list elements must have the same length.

How did you see this? I can just see:

> ══ Failed tests ════════════════════════════════════════════════════════════════
── Error ('test-unit-postprocess.R:5:3'): add_mean_sd.lognormal_samples works with posterior samples from the latent lognormal model ──
Error in `read_cmdstan_csv(files = self$output_files(include_failed = FALSE), 
    variables = variables, sampler_diagnostics = sampler_diagnostics, 
    format = format)`: Assertion on 'files' failed: File does not exist: '/tmp/RtmpVPgfDE/working_dir/Rtmpkb8cQq/model_ea52f49f447bb88d6fe4ace47415d149-202407312111-1-1098f0.csv'.

R/postprocess.R
athowes: Check if returning the same thing and not adding columns.

athowes:
Should we have functionality for getting a "properly" discretised PMF in this package or somewhere else?

R/postprocess.R
athowes: Talked through names F2F and arrived at `predict_delay_parameters` (maybe with alias for `predict_dpars`)

R/postprocess.R
athowes:
```
#' @export
add_mean_sd.default <- function(data, ...) {
  cli::cli_inform(c(
    "!" = "Natural scale mean and standard deviation parameter columns not
    added: no method available for this family",
    "Consider submitting an issue to https:/github.com/epinowcast/epidist"
  ))
  return(data)
}
```

seabbs:
> How did you see this? I can just see:

I just looked through the log?

seabbs:
> Should we have functionality for getting a "properly" discretised PMF in this package or somewhere else?

My vote is somewhere else so that i.e epinowcast can use it and then also use it here but we could dev something here and then port out. Note that `epinowcast` already has a simulate based approach to this but really the approach taken in `EpiAware` (numeric integration) is the way to go I think

R/postprocess.R
```diff
@@ -1,3 +1,137 @@
...
+#' @param fit A model fit with `epidist::epidist`
+#' @inheritParams brms::prepare_predictions
+#' @param ... Additional arguments passed to `brms::prepare_predictions`
+#' @family postprocess
+#' @autoglobal
+#' @export
+predict_delay_parameters <- function(fit, newdata = NULL, ...) {
+  args <- list(...)
+  if ("newdata" %in% names(args)) {
+    newdata <- brms:::validate_newdata(newdata)
```
seabbs, line 12: now that newdata isn't in ... I think it isn't being passed to `brms::prepare_predictions` if this is the case and it wasn't caught by testing this suggests a test is missing we might want to have.

R/postprocess.R
```diff
@@ -1,3 +1,137 @@
...
+#'
+#' @param fit A model fit with `epidist::epidist`
+#' @inheritParams brms::prepare_predictions
+#' @param ... Additional arguments passed to `brms::prepare_predictions`
+#' @family postprocess
+#' @autoglobal
+#' @export
+predict_delay_parameters <- function(fit, newdata = NULL, ...) {
+  args <- list(...)
+  if ("newdata" %in% names(args)) {
```
seabbs, line 11: as new data is an arg we don't need an if check here and it won't be activated regardless. We need to check based on if its NULL I believe

athowes:
> > How did you see this? I can just see:
> 
> I just looked through the log? What was the fix that made this go away?

I didn't do anything it just went away.

seabbs (changes requested):
This is great. I think there are a few errors in the handling of `newdata`. I know we are doing the new data stuff separately but I think we need a few minimal checks here to be sure it works as is and avoid broken code going to main

athowes:
Good spot! Agree we should fix these `newdata` things.

athowes:
I've added tests for this (and indeed they did bring up another code issue!).

R/postprocess.R
athowes: Fixed

R/postprocess.R
athowes: Fixed

seabbs (approved):
Nice. Looks great.

PR #231, opened 2024-08-05
Issue 213: Create functionality for passing in `newdata`

### LINKED ISSUE #213: Help user pass `newdata` as sensible things (e.g. all strata)
In #210 we added functionality to produce predictions (of the delay internal and natural scale parameters) via `brms::prepare_prediction` for any family.

There is an argument `newdata` as follows:

```
An optional data.frame for which to evaluate predictions.
If NULL (default), the original data of the model is used.
NA values within factors are interpreted as if all dummy variables of this factor are zero.
This allows, for instance, to make predictions of the grand mean when using sum coding.
```

Following @seabbs who IMO correctly summarises where we should go:

> We can also provide some functionality to either help users extract unique data points from there data (the simple version of all strata) and potentially to grid expand this to all combinations (i.e observed/unobserved) but I think the first pass should be... (what we have already done)

Basically, here we need to now help users to specify common `newdata` options.

Options as far as I see it are either:

1. Helper functions for them to do that that they call outside `predict_delay_samples`
2. Options in `predict_delay_samples` then put the helper functions inside `predict_delay_samples`

I probably favour 2. over 1. but could be convinced / not strong.

athowes:
## Description

This draft PR will close #213.

## Checklist


newdata.R
```diff
@@ -0,0 +1,30 @@
...
+draws <- posterior::as_draws_df(fit_sex$fit)
+
+# With newdata = NULL
+pred <- predict_delay_parameters(fit_sex)
+
+# With newdata
+strata_df <- prep_obs[1:2, ] |> as.data.frame()
+
+strata_df <- strata_df |>
+  dplyr::select(delay_central, sex, obs_t, pwindow_upr, swindow_upr)
```
athowes, line 18: Maybe could be `NA`?

athowes: Example of use of `newdata` anywhere

athowes: Could also do maths manually (i.e. add the intercept and the covariates then link function)

R/postprocess.R
```diff
@@ -34,6 +34,26 @@ predict_delay_parameters <- function(fit, newdata = NULL, ...) {
...
+#' @autoglobal
+#' @export
+all_strata_newdata <- function(fit) {
+  bterms <- brms::brmsterms(fit$formula)
+  vars <- lapply(bterms$dpars, function(x) all.vars(x$formula))
+  vars <- unique(unlist(vars))
+  var_values <- lapply(vars, function(var) unique(fit$data[, var]))
+  names(var_values) <- vars
+  newdata <- expand.grid(var_values)
+  newdata$delay_central <- 0
```
athowes, line 50: Setting this to any number is the same (add test?)

R/postprocess.R
```diff
@@ -34,6 +34,26 @@ predict_delay_parameters <- function(fit, newdata = NULL, ...) {
...
+  newdata$obs_t <- NA
+  newdata$pwindow_upr <- NA
+  newdata$swindow_upr <- NA
```
athowes, line 53: Extract out LHS of equation bits from `bterms` (make not specific to `latent_individual`)

PR #255, opened 2024-08-26
Issue 235: Constant distributional parameters bug fix

### LINKED ISSUE #235: Vector or real meta bug
We have two issues where the problem of creating a custom Stan function but not knowing if the inputs are `vector` (which is what we have currently) or `real` causes bugs:

* [ ] https://github.com/epinowcast/epidist/issues/227 (bug 2)
* [ ] https://github.com/epinowcast/epidist/issues/222

These are settings where you don't have a model for a particular distributional parameter, and then it's a `real` not a `vector`.

To solve these two issues, we need to check if there is a model on each distributional parameter, and if there isn't then put it to `real` in the custom Stan code rather than `vector`. There also needs to be code to turn that `real` into a `vector` and dispatch on the `latent_lognormal_lpdf` (etc.).

The comments of https://github.com/epinowcast/epidist/issues/222 has more discussion between @seabbs and I on a way to do this. I'm not 100% sure of the route yet, but I think it's doable.

As it's somewhat unlikely a user would not specify a model on a quantity, I'd call this "nice to have" for version 0.1.0 rather than strictly required.

athowes:
## Description

This PR closes #235. It:

* [x] Puts the default `formuala = bf(mu ~ 1)` i.e. removing `sigma ~ 1`
* [x] Changes the Stan code to include either `real` or `vector` in front of the distributional parameters depending on whether there is a model for them. For `mu` is always assumes there is a model. I think this might be a requirement of `brms`, but I'm not sure
* [x] Adds a prior on the `sigma > 0` as the equivalent lognormal to the normal which is then exponentiated if it's an intercept only model
* [x] Doesn't add a test for `bf(mu ~ 1)` working because lots of the tests already have no explicit formula hence will be using this default

## Checklist


seabbs (approved):
Yup this looks good to me! Nice.

seabbs (approved):

PR #259, opened 2024-08-27
Issue 240: Add pointwise log likelihood for latent lognormal model

### LINKED ISSUE #240: Function for pointwise log likelihood
For compatibility of our models with model comparison tools like [`loo`](https://mc-stan.org/loo/) we need to write functions for the pointwise log likelihood.

For example, for the latent lognormal , this would look like the following:

```r
log_lik_latent_lognormal <- function(i, prep) {
  mu <- brms::get_dpar(prep, "mu", i = i)
  sigma <- brms::get_dpar(prep, "sigma", i = i)
  y <- prep$data$Y[i]
  vreal1 <- prep$data$vreal1[i]
  vreal2 <- prep$data$vreal2[i]
  vreal3 <- prep$data$vreal3[i]
  
  # Need to be integrating these out?
  # swindow_raw <- ...
  # pwindow_raw <- ...
  
  # pwindow <- ...
  # swindow <- ... 
  # obs_t <- ...
  # latent_lognormal_lpdf(y, mu, sigma, pwindow, swindow, obs_t)
}
```

Here I have used `latent_lognormal_lpdf` as if it were exported from the Stan code, but it's not possible to do that.

I think that the challenging part of doing this is that the latent variables like `swindow_raw` and `pwindow_raw` need to be integrated out. I could probably write some way to do this using quadrature.

Is this a problem you've already encountered in the academic side of this work @seabbs @kgostic @parksw3? Any pointers?

## How does this fit into the bigger picture

* I think model comparison is an important feature to facilitate, and by writing this one function we get all of the `loo` functionality (I think)
* I think we should demonstrate model comparison in #52
* This issue has been split off from #214 because it's the harder part and not required for prediction

athowes:
## Description

This draft PR will close #240.

## Checklist


seabbs:
I agree we should go for this slightly wrong approach for now.

athowes:
Happy to move forward with slightly wrong approach. I've added an issue to fix it here https://github.com/epinowcast/epidist/issues/267.

Still would like to fix the `wN` point, then I think this is good to go in.

athowes:
So here in the test case:

```r
> str(noverlap)
 int [1:500] 1 2 3 4 5 6 7 8 9 10 ...
> str(woverlap)
 int(0) 
 ```

seabbs (approved):
nice. LGTM

PR #262, opened 2024-08-28
Issue 226: Reconcile Stan and `brms` parameterisation differences

### LINKED ISSUE #226: Stan and `brms` parameterisation issue
In `brms` every family has a `mu` parameter. For example, the gamma distribution parameterisation is `mu` (the mean) and `shape` (the shape parameter).

As such, the fitted `brms` output for a gamma distribution has `Intercept` and `Intercept_shape` parameters.

However, in our `latent_individual` model we create the following likelihood:

```
real latent_gamma_lpdf(vector y, vector mu, vector shape, vector pwindow,
                           vector swindow, array[] real obs_t) {
  int n = num_elements(y);
  vector[n] d = y - pwindow + swindow;
  vector[n] obs_time = to_vector(obs_t) - pwindow;
  return gamma_lpdf(d | mu, shape) - gamma_lcdf(obs_time | mu, shape);
}
```

This likelihood passes `mu` and `shape` to `gamma_lpdf` in Stan. The parameterisation which [`gamma_lpdf`](https://mc-stan.org/docs/2_18/functions-reference/gamma-distribution.html) accepts is `alpha` (shape) and `beta` (inverse rate).

In PR #184 we found that the gamma `latent_individual` model can recover simulation parameters when you take into account that `Intercept` (which you would naively think would be for the mean) is actually for the shape parameter (because that is the first parameter that goes into `gamma_lpdf`.

So we need a solution to this more broadly, not just for the gamma ideally.

`brms` must be doing something internally to handle these transformations, so I think the first step is to understand what it is doing.

athowes:
## Description

This draft PR will close #226.

## Checklist


seabbs:
Plan is to add custom code for gamma right now and circle back to see if we can use the `brms` defaults.

athowes:
My plan here is to first implement in `epidist_stancode.epidist_latent_individual`, add tests, then discuss whether this should be moved into `epidist_family`. Perhaps this PR could just be to solve it then new issue for if it should be moved into `epidist_family`.

seabbs:
Sounds good

athowes:
Added issue on moving this into `epidist_family` or otherwise: https://github.com/epinowcast/epidist/issues/276

seabbs (approved):

PR #273, opened 2024-09-02
Issue 248: Add prediction and log-likelihood methods for the `latent_gamma` family

### LINKED ISSUE #248: Posterior prediction for the gamma family
I think after PR #220 is merged, and then we have a go at #240 then we should try to get these three functions in there for the gamma distribution.

```r
posterior_predict_latent_gamma <- ...
posterior_epred_latent_gamma <- ...
log_lik_latent_gamma <- ...
```

It's possible we may want to have solved the "`brms` and Stan parameterise things differently" issue before we do this, as well (#226).

athowes:
## Description

This PR will close #248.

* [ ] Remains to copy over tests from the `latent_lognormal`
* [ ] Probably want to merge #259 first
* [ ] Then want to add documentation with `latent_lognormal` to those three functions like I have here for `latent_gamma`

## Checklist


R/latent_gamma.R
```diff
@@ -0,0 +1,85 @@
...
+#' @export
+posterior_predict_latent_gamma <- function(i, prep, ...) { # nolint: object_length_linter
+  mu <- brms::get_dpar(prep, "mu", i = i)
+  shape <- brms::get_dpar(prep, "shape", i = i)
+
+  obs_t <- prep$data$vreal1[i]
+  pwindow_width <- prep$data$vreal2[i]
+  swindow_width <- prep$data$vreal3[i]
+
+  .predict <- function(s) {
```
seabbs, line 19: can we take this opportunity to do more with a internal function? (i.e one that takes rgamma as a input)

seabbs (commented):
Looks good so far

R/latent_gamma.R
athowes: Yes, I did think when writing this that it was very repetitive so I agree this is a good idea

seabbs: if you can start to use stuff from `primarycensoreddist` that might save time? If you can't it might suggest ways that package should change

athowes: Nice yeah, no looks like should be able to rely on throughout these functions. Maybe we add this as is working (in order to allow for model comparison lognormal to gamma quick) then make an issue to port over to `primarycensoreddist`?

athowes: Added https://github.com/epinowcast/epidist/issues/275

athowes:
Unit tests failing because need to merge #262 first (wrong parameterisation).

R/latent_gamma.R
```diff
@@ -0,0 +1,85 @@
...
+    d_censored <- obs_t + 1
+    # while loop to impose the truncation
+    while (d_censored > obs_t) {
+      p_latent <- runif(1, 0, 1) * pwindow_width
+      d_latent <- rgamma(1, shape = shape[s], rate = shape[s] / mu[s])
```
seabbs, line 24: very minor but here you parameterise it in terms of the rate and below in terms of the scale. Could you switch so they all use the same parameterization?

tests/testthat/test-unit-latent_gamma.R
```diff
@@ -0,0 +1,76 @@
...
+  draws <- data.frame(obs_t = 1000, pwindow_upr = 0, swindow_upr = 0) |>
+    tidybayes::add_predicted_draws(fit_gamma, ndraws = 100)
+  expect_equal(draws$.draw, 1:100)
+  pred <- draws$.prediction
+  expect_gte(min(pred), 0)
+  expect_true(all(abs(pred - round(pred)) > .Machine$double.eps^0.5))
+})
+
+test_that("posterior_predict_latent_gamma predicts delays for which the data is in the 95% credible interval", { # nolint: line_length_linter.
+  fit_gamma <- readRDS(
```
seabbs, line 35: if this is all we are using it for we could just run a fit in the setup.R file in testthat to get rid of the large saved file

seabbs (changes requested):
This looks good. One very minor style comment

R/latent_gamma.R
athowes: Yes agree this is vestage of me thinking there was a bug, will change

tests/testthat/test-unit-latent_gamma.R
athowes: Yes it is all. This would be annoying in that I run `setup.R` before running tests some of the time but apart from that it's a good solution. I'll add it to the issue on large files

R/latent_gamma.R
athowes: https://github.com/epinowcast/epidist/pull/273/commits/6c0ef8a76b1e285a37e5a3a3776bb7ded8a780cb

tests/testthat/test-unit-latent_gamma.R
athowes: https://github.com/epinowcast/epidist/issues/277#issuecomment-2327120067

seabbs (approved):

PR #282, opened 2024-09-05
Issue 227: Allow setting distributional parameters to constants

### LINKED ISSUE #227: Setting distributional parameters to constant bug
We have two bugs I've seen when trying to set distirbutional parameters to be constant:

```
set.seed(1)

obs_time <- 25
sample_size <- 500

meanlog <- 1.8
sdlog <- 0.5

# The mean of the lognormal distribution is: exp(meanlog + 0.5 * sdlog^2)

sim_obs <- simulate_gillespie() |>
  simulate_secondary(
    dist = rlnorm,
    meanlog = meanlog,
    sdlog = sdlog
  ) |>
  observe_process() |>
  filter_obs_by_obs_time(obs_time = obs_time)

sim_obs <- sim_obs[sample(seq_len(.N), sample_size, replace = FALSE)]
prep_obs <- as_latent_individual(sim_obs)
```

## Bug 1

```
epidist(data = prep_obs, formula = brms::bf(mu ~ 1, sigma = 1), seed = 1)
```

results in:

```
Error: The following priors do not correspond to any model parameter: 
Intercept_sigma ~ normal(0, 0.5)
Function 'default_prior' might be helpful to you.
```

**Reason**: in `epidist_prior` we hard code that the model has to have certain priors.

## Bug 2

Even if we didn't have this prior problem, then we have the following bug:

```
epidist(data = prep_obs, formula = brms::bf(mu ~ 1, shape = 1), family = stats::Gamma(link = "log"))
```

```
Compiling Stan program...
Semantic error in '/var/folders/4r/hkp4v9fn3wx044_hk8qnsjxw0000gn/T/RtmpSfhylK/model-3c965f0b3c81.stan', line 61, column 14 to column 72:
   -------------------------------------------------
    59:      mu += Intercept;
    60:      mu = exp(mu);
    61:      target += latent_gamma_lpdf(Y | mu, shape, pwindow, swindow, vreal1);
                       ^
    62:    }
    63:    // priors including constants
   -------------------------------------------------

Ill-typed arguments supplied to function 'latent_gamma_lpdf':
(vector, vector, real, vector, vector, array[] real)
Available signatures:
(vector, vector, vector, vector, vector, array[] real) => real
  The third argument must be vector but got real
```

**Reason**: our custom likelihood expects a vector for all distributional parameters. If instead its a real then it throws an error.

athowes:
## Description

This PR closes #227.

## Checklist

### DIFF
--- a/R/prior.R
+++ b/R/prior.R
@@ -90,12 +90,20 @@ epidist_family_prior.default <- function(family, formula, ...) {
 epidist_family_prior.lognormal <- function(family, formula, ...) {
   prior <- brms::prior("normal(1, 1)", class = "Intercept")
   if ("sigma" %in% names(formula$pforms)) {
-    prior <- prior +
-      brms::prior("normal(-0.7, 0.4)", class = "Intercept", dpar = "sigma")
+    # Case with a model on sigma
+    sigma_prior <- brms::prior(
+      "normal(-0.7, 0.4)", class = "Intercept", dpar = "sigma"
+    )
+  } else if ("sigma" %in% names(formula$pfix)) {
+    # Case with sigma fixed to a constant
+    sigma_prior <- NULL
   } else {
-    prior <- prior +
-      brms::prior("lognormal(-0.7, 0.4)", class = "sigma", lb = 0, ub = "NA")
+    # Case with no model on sigma
+    sigma_prior <- brms::prior(
+      "lognormal(-0.7, 0.4)", class = "sigma", lb = 0, ub = "NA"
+    )
   }
+  prior <- prior + sigma_prior
   prior$source <- "family"
   prior[is.na(prior)] <- "" # This is because brms likes empty over NA
   prior[prior == "NA"] <- NA # To keep particular NA
--- a/R/diagnostics.R
+++ b/R/diagnostics.R
@@ -40,7 +40,7 @@ epidist_diagnostics <- function(fit) {
     diagnostics <- data.table(
       "time" = sum(rstan::get_elapsed_time(fit$fit)),
       "samples" = nrow(np) / length(unique(np$Parameter)),
-      "max_rhat" = round(max(brms::rhat(fit)), 3),
+      "max_rhat" = round(max(brms::rhat(fit), na.rm = TRUE), 3),
       "divergent_transitions" = sum(np[divergent_indices, ]$Value),
       "per_divergent_transitions" = mean(np[divergent_indices, ]$Value),
       "max_treedepth" = max(np[treedepth_indices, ]$Value)
--- a/tests/testthat/test-int-latent_individual.R
+++ b/tests/testthat/test-int-latent_individual.R
@@ -81,6 +81,22 @@ test_that("epidist.epidist_latent_individual fits and the MCMC converges in the 
   expect_convergence(fit)
 })
 
+test_that("epidist.epidist_latent_individual fit and the MCMC converges when setting sigma = 1 (a constant)", { # nolint: line_length_linter.
+  # Note: this test is stochastic. See note at the top of this script
+  skip_on_cran()
+  set.seed(1)
+  fit_constant <- epidist(
+    data = prep_obs,
+    formula = brms::bf(mu ~ 1, sigma = 1),
+    seed = 1,
+    silent = 2,
+    output_dir = fs::dir_create(tempfile())
+  )
+  expect_s3_class(fit_constant, "brmsfit")
+  expect_s3_class(fit_constant, "epidist_fit")
+  expect_convergence(fit_constant)
+})
+
 test_that("epidist.epidist_latent_individual recovers the simulation settings for the delay distribution in the default case", { # nolint: line_length_linter.
   # Note: this test is stochastic. See note at the top of this script
   skip_on_cran()


athowes:
Sorry I think there are thing I can fix with the CI, shouldn't have tagged for review!

athowes:
If you have a `sigma = 1` then the Rhat value will be `NA` for that parameter (which was causing convergence checking to fail). I have set the maximum Rhat now to remove `NA` entries

R/prior.R
```diff
@@ -90,12 +90,20 @@ epidist_family_prior.default <- function(family, formula, ...) {
 epidist_family_prior.lognormal <- function(family, formula, ...) {
   prior <- brms::prior("normal(1, 1)", class = "Intercept")
   if ("sigma" %in% names(formula$pforms)) {
-    prior <- prior +
-      brms::prior("normal(-0.7, 0.4)", class = "Intercept", dpar = "sigma")
+    # Case with a model on sigma
+    sigma_prior <- brms::prior(
+      "normal(-0.7, 0.4)", class = "Intercept", dpar = "sigma"
+    )
+  } else if ("sigma" %in% names(formula$pfix)) {
```
seabbs, line 97: Why did you choose to do this here on a case by case basis and not in epidist_prior (i.e look for anything in the formula that is fixed and remove it from the currently specified prior?

R/prior.R
```diff
@@ -90,12 +90,20 @@ epidist_family_prior.default <- function(family, formula, ...) {
...
+    # Case with sigma fixed to a constant
+    sigma_prior <- NULL
   } else {
-    prior <- prior +
-      brms::prior("lognormal(-0.7, 0.4)", class = "sigma", lb = 0, ub = "NA")
+    # Case with no model on sigma
+    sigma_prior <- brms::prior(
```
seabbs, line 102: not for now but it seems like the solution above could be applied here quite easily if you also knew the link (at least in the log case).

athowes: Solution above as in putting things into `epidist_prior`?

seabbs: yes and trying to automate vs on a case by case basis

R/prior.R
athowes: `replace_prior` currently errors when you pass it a prior which there is no matching prior for it to overwrite. One possibility would be just to make it less strict and just to have a warning when it's given a prior which doesn't exist. But this should only happen for the user passing in priors, of course for all our code we shouldn't have routine warnings happening.

To remove anything which is fixed it would be like removing anything which is fixed from each of:

```r
model <- epidist_model_prior(data, formula)
family <-  epidist_family_prior(family, formula)
```

I can imagine the code for this being like:

```r
model <- epidist_model_prior(data, formula)
family <-  epidist_family_prior(family, formula)
for(dpar in names(formula$pfix)) { ... }
```

Which I could be happy to move to. It's probably more complicated now but I agree in future it'll be less complicated. It's also mildly unappealing to have e.g. `epidist_model_prior` return the wrong prior (all else equal then it being lower down the stack is better).

R/prior.R
athowes: Yeah I agree about trying to automate this as we add more priors

R/prior.R
seabbs: > One possibility would be just to make it less strict and just to have a warning when it's given a prior which doesn't exist. But this should only happen for the user passing in priors, of course for all our code we shouldn't have routine warnings happening.

I like this suggestion and then run is it constant checks on the complete output as another internal function.

If you feel like this is to much for this PR we can merge and open a new issue but I think we want this generalised ASAP vs pushing down the priority order.

> epidist_model_prior return the wrong prior (all else equal then it being lower down the stack is better).

We don't really expect users to interact with this at all do we so in some sense it doesn't matter?

athowes: > If you feel like this is to much for this PR we can merge and open a new issue but I think we want this generalised ASAP vs pushing down the priority order.

Yes I feel like a solution along these lines could work but it's not obvious so I'll make a new issue.

> We don't really expect users to interact with this at all do we so in some sense it doesn't matter?

Developers are still a type of user.

R/prior.R
athowes: https://github.com/epinowcast/epidist/issues/305

R/prior.R
seabbs: do you think developers will want to write endless if else statements in their code over it having output that is overwritten ;)

athowes: Yes it makes me feel productive.

R/prior.R
seabbs: I think this needs its own issue as its a slightly different problem to the no model question (though related) and may not need to be dealt with in its own PR.

R/prior.R
seabbs: #305

seabbs: #305

athowes: https://github.com/epinowcast/epidist/issues/306

seabbs (approved):
Works for now but I think we want to push on #305 as this is increasingly brittle. This can be merged once we have a specific issue for generalisng the handling of intercepts on sigma as we have discussed a range of different options there which I think would resolve (i.e something like #305 or instead modifying the formula so we always have sigma ~ 1 vs supporting sigma with no intercept as well).

athowes:
> This can be merged once we have a specific issue for generalisng the handling of intercepts on sigma as we have discussed a range of different options there which I think would resolve (i.e something like https://github.com/epinowcast/epidist/issues/305 or instead modifying the formula so we always have sigma ~ 1 vs supporting sigma with no intercept as well).

I've added mentioning that `~ 1` would solve #305 as well as doing something around automatic changes to the prior. Is this sufficient? I'd prefer not to have another issue for this.

athowes:
I've added a meta issue about it: https://github.com/epinowcast/epidist/issues/307

R/prior.R
athowes: https://github.com/epinowcast/epidist/issues/307

PR #300, opened 2024-09-10
Issue 79: Make some functions internal

### LINKED ISSUE #79: Decide on which functions to make internal
Currently (approximately) all of the functions in the package are external (`@export`).

We should decide which should be made internal to the package.

As example, in #70 @seabbs raised this regarding functions in `utils.R` and the `epidist_stancode` functions.

athowes:
## Description

This draft PR closes #79.

It is my proposal for which functions to be made internal:

* Any `utils.R`
* Any custom `brms` family prediction and log likelihood functions
* I also moved `replace_prior` into `utils.R`

I used `#' @noRd`. Another alternative would be to use `#' @keywords internal`?

## Checklist


seabbs:
> #' @keywords internal

I would prefer this

R/utils.R
```diff
@@ -39,8 +37,53 @@ epidist_version_stanvar <- function() {
...
+#' This function takes `old_prior` and replaces any prior distributions
+#' contained in it by the corresponding prior distribution in `new_prior`.
+#' If there is a prior distribution in `new_prior` with no match in `old_prior`
+#' then the function will error and give the name of the new prior distribution
+#' with no match.
+#'
+#' @param old_prior One or more prior distributions in the class `brmsprior`
+#' @param new_prior One or more prior distributions in the class `brmsprior`
+#' @importFrom cli cli_inform
+#' @importFrom utils capture.output
```
seabbs, line 57: yourt missing the dplyr functions from importFrom

athowes:
> I would prefer this

Why?

R/utils.R
```diff
@@ -6,8 +6,7 @@
 #' @param path The path within the `stan/` folder of the installed `epidist`
 #' package to the Stan code chunk of interest.
 #' @return A character string containing the Stan code chunk of interest.
-#' @family utils
-#' @export
+#' @noRd
```
seabbs, line 9: for all internal functions we might consider a rename. 1. They likely don't need the epidist prefix and 2. we might want to use common these are internal signals like . or _ as a prefix

seabbs (commented):
I agree with all the suggestions to go internal. Some specific points about the process.

seabbs:
> Why?

because then they are still doc'd if people want to know how they work. I think that the docs for noRd have a warning i.e do you really want to do this as I think the default is that they are internal but still have a help fie

R/utils.R
athowes: Happy with remove `epidist_`. Don't have major desire to do `.` or `_` but if there is a push I can do it

seabbs: The reason is to indicate these are internal functions vs not. Its generally pretty standard practice so I would favour doing so but I don't feel strongly

athowes: For the `brms` ones I think that would break things

athowes: [a3447be](https://github.com/epinowcast/epidist/pull/300/commits/a3447be829f7a91d62270020d08925bdbb1c7821) does the remove `epidist_` part

R/utils.R
athowes: [19104b3](https://github.com/epinowcast/epidist/pull/300/commits/19104b3f08cce359d07c2811815ed17bc8b2bd15)

R/utils.R
seabbs: > For the brms ones I think that would break things

Then don't do it for those ones. What is the push back for the others?

athowes: It's just not something I'm familiar with standard practise on / I have some doubts about the extent to which it is standard practise. But OK as I say I don't mind so I'll add it

athowes: "Have some doubts" --> pick an R package I think they would know what they are doing, then what proportion of the time are they doing this? For example I picked `ggplot2` and I don't see it there

pearsonca: can't start with underscores - that's just a weird bit of R naming rules (https://cran.r-project.org/doc/FAQ/R-FAQ.html#What-are-valid-names_003f). I've definitely seen the starts-with-dot for internals.

pearsonca: (it's probably leaning into the base R `.Internal()` as an emphasis idiom)

athowes: I think done in [d45d0c9](https://github.com/epinowcast/epidist/pull/300/commits/d45d0c9afdd08711badba3e6212d6680706c05ab)

athowes:
Yeah probably, is there a good way to do this rebase? When I do `pull origin main` on the branch then I have all these docs conflicts but seems bad to be going through and hand fixing them. Want to pull the non-docs then run `document`

seabbs:
> Edit: I tried the second option, keep all conflicts then overwrite.

Sorry yes this is what I would do

seabbs (approved):

PR #316, opened 2024-09-11
Issue 314: Cut functions

athowes:
## Description

This PR closes #314.

## Checklist

- [ ] My PR is based on a package issue and I have explicitly linked it.
- [ ] I have included the target issue or issues in the PR title in the for Issue(s) *issue-numbers*: PR title
- [ ] I have read the [contribution guidelines](https://github.com/epinowcast/.github/blob/main/CONTRIBUTING.md).
- [ ] I have tested my changes locally.
- [ ] I have added or updated unit tests where necessary.
- [ ] I have updated the documentation if required.
- [ ] My code follows the established coding standards.
- [ ] I have added a news item linked to this PR.
- [ ] I have reviewed CI checks for this PR and addressed them as far as I am able.

PR #328, opened 2024-09-13
Issue #304: Update approximate inference vignette and remove `draws_to_long`

### LINKED ISSUE #241: Expected delay variance using `tidybayes`
The `tidybayes` package via `brms` has functionality to get the expected prediction `.epred`.

What we might also like to do is to get the expected prediction variance. This is implemented in `add_mean_sd` but perhaps there is a more native way to do it.

### LINKED ISSUE #304: Use `tidybayes` in approximate inference vignette
The approximate inference vignette still uses old post-processing functionality (that maybe we should delete). It should be updated to use `tidybayes` for post-processing and figures.

athowes:
## Description

This PR closes #304.

1. `tidybayes` can't yet do the functionality in the approximate inference vignette because it doesn't have the SD. See issue #241.
2. As a result I didn't get rid of `predict_delay_parameters` yet
3. But there were other things to fix about the approximate inference vignette including making the posterior figure better and adding figure captions
4. I also removed use of `draws_to_long` in the vignette, then everywhere, then deleted the function. This is inline with our plan to cut down on features that aren't really used (e.g. #316)

Working on this was a reminder that we need to resolve #241 so that we can complete axing the `predict_delay_parameters` function. I'll make an issue along those lines.

## Checklist


vignettes/approx-inference.Rmd
```diff
@@ -171,7 +172,12 @@ fits <- list(
 
 draws <- imap(fits, function(fit, name) {
   predict_delay_parameters(fit) |>
-    draws_to_long() |>
+    as.data.frame() |>
```
seabbs, line 175: are you sure you need this as.data.frame call? It doesn't matter much as will go with a refactor to dplyr but might as well not include here if don't need.

seabbs (dismissed):
LGTM. See note about as.data.frame conversion

vignettes/approx-inference.Rmd
athowes: Yeah I think we need it for some reason (and that's bad)

athowes:
I think that this will pass all tests apart from R CMD CHECK as Windows latest.

That being the case, I think we should merge it, as that is the same issue that we have on main at the moment and it's not related to this PR.

I've added an issue #332 where I'll try to fix this.

seabbs (approved):
LGTM. Going to bypass checks as reported failing is happening on main.

Issue #307, opened 2024-09-11
No model meta


athowes:
In the case that there is no model on a parameter we can either

1. Work around it by setting appropriate priors (#305)
2. Enforce there being a model by putting `~ 1` internally (https://github.com/epinowcast/epidist/issues/266)

This is a meta issue for deciding between these options, or suggesting new ways to handle this.

seabbs:
I strongly think we should do 2 for simplicity

kgostic:
Agreed in meeting that we're doing 2.

athowes:
* Also shift to `brms` style formula input.
* If given string then add on all other formula as constant
* Require formula on everything if given `brms::bf()`
* Have to also deal with the fixing stuff to a constant case

seabbs:
sounds good

athowes:
> `bmr(mu ~ 1 +w wefwef)`

What's this?

seabbs:
yes exactly sorry should have been clearer

athowes:
Closing as outdated / no longer relevant.

Issue #306, opened 2024-09-11
Generalise handling priors for parameters set to constant


athowes:
https://github.com/epinowcast/epidist/pull/282#discussion_r1753960842 (and comment thread):

> `replace_prior` currently errors when you pass it a prior which there is no matching prior for it to overwrite. One possibility would be just to make it less strict and just to have a warning when it's given a prior which doesn't exist. But this should only happen for the user passing in priors, of course for all our code we shouldn't have routine warnings happening.
> 
> To remove anything which is fixed it would be like removing anything which is fixed from each of:
> 
> ```r
> model <- epidist_model_prior(data, formula)
> family <-  epidist_family_prior(family, formula)
> ```
> 
> I can imagine the code for this being like:
> 
> ```r
> model <- epidist_model_prior(data, formula)
> family <-  epidist_family_prior(family, formula)
> for(dpar in names(formula$pfix)) { ... }
> ```
> 
> Which I could be happy to move to. It's probably more complicated now but I agree in future it'll be less complicated. It's also mildly unappealing to have e.g. `epidist_model_prior` return the wrong prior (all else equal then it being lower down the stack is better).

One plan here:

1. Change `replace_prior` to do a warning rather than error when there is no prior to replace
2. Add utility function to remove all parameters set to a constant
3. Put use of that utility function into `epidist_prior` and remove the lower level parts about removing priors on constants

The flaws in this plan:

1. If we change `replace_prior` to just do a warning then we don't need the utility function to remove priors on constant parameters since `default_prior` already won't have put a prior on a constant in there, so there will be nothing to replace, hence we just end up with a warning
2. I don't think users should have to look at warnings for things we are doing internally, the warnings should just come up if they place a prior that doesn't exist

athowes:
Possible solution:

1. Have a version of `replace_prior` which doesn't care if you try to replace things that don't exist and use that for internal replacements.
2. Also have a version which gives warnings and use that for user replacements.

Downside: maybe it's harder for us to be debugging if any replacements are going wrong.

seabbs:
just have an optional arg that you set when its time to replace user priors I think

athowes:
As a part of this issue, we might like to also alter the structure of our priors so that `epidist_family_prior.lognormal` doesn't need as much general code in it.

PR #370, opened 2024-10-03
Issue #266: Include intercept when no model provided

### LINKED ISSUE #266: Consider setting intercept on any distributional parameters without a model
In PR #255 we moved to allowing `bf(mu ~ 1)` to be the default for the `formula` option of `epidist()`.

This means that users now do not need to specify a model on the distributional parameters of the model (aside from `mu`) if they don't want to.

Although this is nice, it does now mean that:

1. We must be specifying priors on both the `Intercept_dpar` and `dpar` parameters in `epidist_family_prior()`
2. If we have any post-processing which relies on knowing model names (e.g. `Intercept_dpar`) then it'll be more fiddly

An alternative proposal is to set `dpar ~ 1` for any `dpar` left without a model. This would fix the two problems above but introduce its own problems:

1. Users may be confused that we've changed their model in the output
2. If a user puts a prior on `dpar` as an override then it won't easily work with the new `Intercept_dpar` parameter. Further: perhaps it's good to allow users to be putting priors on the distributional parameter itself, rather than the intercept

For the time being I'm happy to stick with things as they are, but if the downsides become more painful for some reason I could be convinced to switch approaches.

athowes:
## Description

This PR closes #266.

It:

* Adds a utility function `.make_intercepts_explicit` which takes a `formula` and puts a model `dpar ~ 1` on any distributional parameters which 1. are not fixed, 2. are not `mu`, 3. do not have a formula
* This utility function is used in `epidist_formula.epidist_latent_individual`
  * Note that with https://github.com/epinowcast/epidist/issues/372 we should move this out of being model-specific
* Simplifies `epidist_family_prior.lognormal` because no longer have to set priors on the untransformed `sigma`
* Requires follow up work to:
  * Simplify `epidist_family_prior.lognormal` with restructure. Probably this fits into https://github.com/epinowcast/epidist/issues/306
  * Add new issue on shifting to shifting to `brms` style formula input rather than require `brms::bf()` as discussed in #307
* Also fixed a bug in test that `brms` samples from the prior right which was highlighted by these changes [9d19d49](https://github.com/epinowcast/epidist/pull/370/commits/9d19d4950ee51177b81387397a31c30541ce63b6) (not sure how it was passing before, perhaps because there wasn't an explicit model it worked OK)

## Checklist


R/latent_individual.R
```diff
@@ -155,6 +155,11 @@ epidist_formula.epidist_latent_individual <- function(data, family, formula,
   epidist_validate(data)
   formula <- brms:::validate_formula(formula, data = data, family = family)
 
+  # Here could detect which dpar do not have a formula and change it so that
+  # they do. Obviously this isn't a model-specific task so perhaps it's
+  # suggesteing that we want to be generalising epidist_formula in some way
+  # Or it could be a helper function and we put it in all of the epidist_formula
+  
   formula <- stats::update(
```
athowes, line 163: This is the only `latent_individual` specific part of `epidist_formula.latent_individual`.

Create a new internal S3 which does this and make the rest of it a regular function. `as_epidist_formula`, `convert_formula_to_epidist_model`... ask ChatGPT.

athowes: #372

seabbs (approved):
Nice this is great. All follow up tasks look good to me.

Issue #33, opened 2024-05-14
Refactor `observe_process` function


athowes:
## New plan

* Create `add_event_vars` function to add upper and lower given either lower and window or lower and upper
* Enforce our names structure which is `ptime_lwr` and `ptime_upr` (i.e. change the names)
* Find a way to slot the back into existing functionality so as not to break anything for now

## Old

I've just used `observe_process()`. Some notes on possible refactor:

* The name "observe process" is vague and seems misleading for what it is doing. For my understanding, what it is doing is implementing interval censoring. If so, the name should reflect that.
* If it is implementing interval censoring, then I think that there should be arguments available for specifying the censoring intervals. At the moment it looks like it's just using daily intervals, and some kind of fixed upper and lower.

seabbs:
I think its implementing the observation process which in this instance is censoring but agree it can be renamed.

athowes:
Even in this instance I'd think of the observation process as having multiple stages, including but not limited to censoring. I think having functions to modularly implement each part of the observation process would be nice.

seabbs:
Yes agree

athowes:
As a part of this issue, I propose to:

* [ ] rename `observe_process()` to something to represent interval censoring,
* [ ] include arguements for the intervals to be used,
* [ ] create unit tests for the function,
* [ ] fill out the `roxygen2` documentation for the function.

athowes:
> As a part of this issue, I propose to:
> 
>     * [ ]  rename `observe_process()` to something to represent interval censoring,
> 
>     * [ ]  include arguements for the intervals to be used,
> 
>     * [ ]  create unit tests for the function,
> 
>     * [ ]  fill out the `roxygen2` documentation for the function.

I am now working in this direction.

athowes:
Note on implementation:

* I think we should be using `.floor_mult` here to round down to the nearest multiple of `x`
* I guess there is a question about whether people are censoring their data or they are taking already censored data and adding some columns. Maybe it just works the same either way...
* I think we should be letting users input `x`
* Probably we can let there be different censoring windows used for primary and secondary
  * Then I think we can re-use the names `pwindow` and `swindow` for the sizes of these
* I don't know why we have `_daily` as new columns. If I just delete these does it do anything? Use code search to find all instances of `_daily`

athowes:
I've drafted a function that would look like this:

```r
observe_process <- function(linelist, pwindow = 1, swindow = 1) {
[...]

PR #375, opened 2024-10-10
Issue #33: Write `add_event_vars` preprocessing functionality

athowes:
## Description

This PR will close issue #33. The goal is to add a modular function which adds interval censoring columns.

* Adds a function `add_event_vars`
* This function:
  1. Gets the columns of `linelist` renamed into our preferred names
  2. Allows users to pass `pwindow` and `swindow` as either charater point to the name of the column, or numeric in which case the column will be created
  3. Creates the `lwr` `upr` and `window` columns from other columns if names pointing to them not provided 
  4. Checks that the columns are numeric, above zero, and add up to equal each other when relevant
  5. Rearranges the columns
* Adds a helper function `.rename_column` which renames the column if provided two strings and the dataframe
* This function does not yet slot into anywhere within the code. I have left `observe_process()` alone for now. Eventually this function plus others will slot in to replace it

About reviewing this PR:

* I think I might have made the function a bit overly complicated so any help reducing the complexity appreciated
* There are also some cases given the amount of complexity where I am not 100% sure it does what we want

Remains to do:

* Document `add_event_vars` better (i.e. the parameters have no description at the moment)
* Document `.rename_column` and maybe add tests for it

## Checklist


seabbs:
> I think I might have made the function a bit overly complicated so any help reducing the complexity appreciated

I haven't read the code yes but agree from reading the overview. My current uninformed view is that this is really a few functions combined. 

I think some clarity here could be had in thinking about what exactly is happening here and giving that names

R/preprocess.R
```diff
@@ -0,0 +1,69 @@
...
+    linelist$pwindow <- pwindow
+  }
+
+  if (is.numeric(swindow)) {
+    cli::cli_warn("Overwriting using numeric value(s) of swindow provided!")
+    linelist$swindow <- swindow
+  }
+
+  if (is.null(stime_upr)) {
+    linelist <- mutate(linelist, stime_upr = stime_lwr + swindow)
```
seabbs, line 35: Performance doesn't matter hugely but noting as written this is a very inefficient function as every rename, mutate, etc all here leads to a data copy.

R/preprocess.R
```diff
@@ -0,0 +1,69 @@
...
+#' @param ptime_upr ...
+#' @param pwindow ...
+#' @param stime_lwr ...
+#' @param stime_upr ...
+#' @param swindow ...
+#' @family preprocess
+#' @autoglobal
+#' @export
+add_event_vars <- function(
+  linelist, ptime_lwr = NULL, ptime_upr = NULL, pwindow = NULL,
```
seabbs, line 14: I think adding event vars is really going from having p time in some format and s time in some format (lower bound assumed) and adding pwindow and window to get the event variables. 

I think the step of making sure we have _lwr and upr for each is another function (and would perhaps naturally live in a general as_epidist_linelist class which can then optionally be passed to as_latent_model. 

If not that s3 design then something else.

seabbs (commented):
I added some thoughts after looking at the code. As I said I think this function is trying to do too much. One option I have suggested is an initial s3 class as_epidist_linelist we can then convert into model formats which might be nice. 

A path here is we accept this PR nearly as is and write a new issue to clean up. Another option is we simplify the functionality here and address the rest later.

athowes:
Yep agree want more clarity here.

I was also looking at `as_latent_individual` and I think even though we can't replace `observe_process` we might be able to start getting rid of bits of `as_latent_individual` within these PRs.

seabbs (approved):
Accepting under the assumption is will shortly be refactored. Noting that the partner issue is missing detail of the pr conversation here

PR #379, opened 2024-10-13
Issue #306: Simplify prior infrastructure by allowing warning when overwriting

athowes:
## Description

This PR will close Issue #306. The goal is to prevent having to specify what to do (not set a prior) if there is a constant set on a distributional parameter.

* Adds `abort` option to `.replace_prior` which by default is `FALSE` and instead a warning will be provided
  * Arguably we don't need the warning i.e. if running with `sigma = 1` at the end you get the warning that:
 
```r
Warning message:
ℹ No available prior to replace in old_prior found for:
Intercept_sigma ~ normal(-0.7, 0.4) 
``` 
 
* Uses the `abort` option only for user input priors in `epidist_prior`
  * Arguably we could not do this and just let `brms` catch it
  * Users might be passing priors which don't work / which are just ignored by `brms`. But perhaps fine with a warning rather than an abort?
* Gets rid of the "setting a `NULL` prior if parameter set to a constant" in `epidist_family_prior.lognormal`

So overall possibly we want to move to ignore / warn rather than warn / abort as the two options. I do like knowing that I get a warning as a developer though if I do something wrong (though the warnings all the time about the constant are annoying).

## Checklist


R/utils.R
```diff
@@ -52,9 +52,12 @@
 #'
 #' @param old_prior One or more prior distributions in the class `brmsprior`
 #' @param new_prior One or more prior distributions in the class `brmsprior`
+#' @param abort If `TRUE` then if a `new_prior` is provided for which there is
```
seabbs, line 55: Some thoughts: 

1. `epidist_prior` could take an argument about how strict to be
2. This "prior_present" could be a mulit-option string arg (i.e "strict", "warn" and "allow-mismatch").
3. I think my preference is a no warning as I can see this being really annoying for users with only very occasionally benefits.

seabbs (commented):
This looks good. I added some thoughts on making it more user settable and less noisy by default

R/utils.R
athowes: Note that users don't call `epidist_prior` it's internal so we'd need a pass through option in `epidist` to allow it to be settable

seabbs: they could call it though right? Its exported and just takes common  arguments?

seabbs: if yes to above we could doc that investigation path or as you say we could pass through the arg.

athowes: Yep let me investigate this. I did consider if there is some general outer "options" input that we want (like a general `verbose` or `strict` option)

athowes:
Current status is:

* If a user puts a constant on a parameter then we will try to overwrite it with a prior and there will be a warning
* If a user puts a prior on something that doesn't exist then we will abort it

Ideal behaviour:

* We do not try to put a prior when there is a constant, but this requires us figuring things out
* Users are warned about them putting priors on things which don't exist (but not in an overly annoying way)

R/utils.R
athowes: I do not think we need something as sophisticated as multistring option here. Proposed solution in comments below (and code).

R/utils.R
```diff
@@ -67,15 +69,16 @@
...
+        select(prior = prior_new, dplyr::all_of(cols), source = source_new)
     ))
-    msg <- c(
-      "i" = "No available prior to replace in old_prior found for:",
-      missing_prior
-    )
-    cli_abort(message = msg)
+    if (warn) {
+      msg <- c(
+        "!" = "No available prior to replace in old_prior found for:",
```
seabbs, line 76: From the user perspective they are passing in prior so I think it would be better to update the internal args to respect this and the warning to match

seabbs (approved):
This looks reasonable I think if my understand is correct that the test shows that updating a prior when the user passes a constant is silent.

athowes:
> This looks reasonable I think if my understand is correct that the test shows that updating a prior when the user passes a constant is silent.

When the users passes a constant then yes it is silent. If they pass a constant and a prior on that constant then no it is not silent. Summary: anything we pass is silent, anything they pass and isn't suitable has a warning.

R/utils.R
athowes: Yes fair enough I think there should be a way to rewrite this function to make more sense from external perspective (even though it is an internal function and makes sense as is now)

seabbs (approved):

PR #383, opened 2024-10-14
Issue #372: Refactor formula dispatch

### LINKED ISSUE #372: Refactor of `epidist_formula` to use S3
> This is the only `latent_individual` specific part of `epidist_formula.latent_individual`.
> Create a new internal S3 which does this and make the rest of it a regular function. `as_epidist_formula`, `convert_formula_to_epidist_model`... ask ChatGPT.

_Originally posted by @athowes in https://github.com/epinowcast/epidist/pull/370#discussion_r1786027060_

In PR #338 we added double dispatch to `epidist_family`. I think similarly that we should refactor `epidist_formula` so that any non-model-specific parts are not containing within model-specific functions. I don't think we need double dispatch, just single dispatch and an outer function.

```r
epidist_formula <- function () {
  # things here
  # including the helper in PR #370
  epidist_formula_model()
  # ...
}
```

Unless there are any formula specific bits? I don't think there would be.

### LINKED ISSUE #382: Reorganise function locations
At the moment we have functions stored in files like:

* `generics.R`
* `defaults.R`

These were related to S3 methods.

I think that we should reorganise so that functions are grouped together by functionality rather than by if they are generics or methods. So I'd advocate for moving these out into files like:

* `formula.R`
* `family.R`
* `fit.R`

To some extent this has already started to happen (for example there is already `family.R`).

An auxiliary goal of this is that if we want to add a new model or a new family for an existing model it should be one new script.

As a part of this issue the function groupings on the `pkgdown` also needs to change.

athowes:
## Description

This draft PR will close #372.

I am considering doing #382 as well since #372 involves reorganising the function locations. In doing so it seems easy to also fix the other function locations. I'd like to move quickly.

What has been done:

On #372:

* `epidist_formula` is now a regular function which contains all processing that happens across models
* It calls `epidist_formula_model` which dispatches on the class of `data` into the model specific part
* Moved into new file `formula.R` and the tests are moved into a `test-formula.R` file

On #382:

* Moves formula functions into `formula.R`

Edit: I'm going to let this get to passing tests and if it's quick and I can catch a reviewer I'll put the #382 into a new PR.

## Checklist


R/formula.R
```diff
@@ -0,0 +1,48 @@
...
+  UseMethod("epidist_formula_model")
+}
+
+#' Default method for defining a model specific formula
+#'
+#' @inheritParams epidist_formula
+#' @param ... Additional arguments passed to method.
+#' @family formula
+#' @export
+epidist_formula_model.default <- function(data, ...) {
```
seabbs, line 43: I think the default should be just to return the formula in order to allow for models that don't require formula changes not to have to write boiler plate they don't need (this is true anywhere else we run into this I think).

seabbs (commented):
Really nice. One small comment about opening up the default method now that this is optional.

R/formula.R
seabbs: yup I think it did make sense but as you say things have changed.

athowes:
Actually we have the same issue in `family`:

```r
#' Default method for defining a model specific family
#'
#' @inheritParams epidist_family_model
#' @param ... Additional arguments passed to method.
#' @family family
#' @export
epidist_family_model.default <- function(data, ...) {
  cli_abort(
    "No epidist_family_model method implemented for the class ", class(data),
    "\n", "See methods(epidist_family_model) for available methods"
  )
}
```

Can fix both here.

seabbs (approved):
Nice. LGTM

PR #389, opened 2024-10-15
Issue #382: Reorganise function locations

### LINKED ISSUE #382: Reorganise function locations

athowes:
## Description

This PR will close #382. Creates:

* `fit.R`
* `stancode.R`
* `validate.R`

Removes:

* `defaults.R`
* `generics.R`

I think this is much clearer as a structure.

Follow up issues to create or add comments to existing issues:

* [x] `epidist_stancode.default` can just pass back `NULL` or whatever it is does nothing. Doesn't need to abort.
* [ ] `epidist_validate.default` contains the catcher for data without a model class. Could make this error message more informative then?
* [ ] Do we want to export validate functions? Stan code functions? Users are not going to need to create these functions but they will need to create instances of the methods for their model

## Checklist


seabbs (approved):
Nice looks good

seabbs:
Agree with the points for further issues. 

> Do we want to export validate functions? Stan code functions? Users are not going to need to create these functions but they will need to create instances of the methods for their model

IF we move to the validation model that scoring utils now uses we may be able to side step this discussion (validation .[ methods)

seabbs:
Agree with the points for further issues. 

> Do we want to export validate functions? Stan code functions? Users are not going to need to create these functions but they will need to create instances of the methods for their model

IF we move to the validation model that scoring utils now uses we may be able to side step this discussion (validating.[ methods). As it stands I think we need to export it

seabbs (approved):

PR #396, opened 2024-10-18
Issue #374: Set default to `mu ~ 1` as formula

athowes:
## Description

This PR closes #374.

I think we already had things set up to work fine with `formula = mu ~ 1` because `epidist_formula` is calling:

```r
formula <- brms:::validate_formula(formula, data = data, family = family)
```

So it already is going to convert from `mu ~ 1` to ` brms::bf`.

I added one unit test to check this.

Edit: this unit test was later removed in favour of using `mu ~ 1` as the default.

## Checklist

- [x] My PR is based on a package issue and I have explicitly linked it.
- [x] I have included the target issue or issues in the PR title in the for Issue(s) *issue-numbers*: PR title
- [x] I have read the [contribution guidelines](https://github.com/epinowcast/.github/blob/main/CONTRIBUTING.md).
- [x] I have tested my changes locally.
- [x] I have added or updated unit tests where necessary.
- [x] I have updated the documentation if required.
- [x] My code follows the established coding standards.
- [ ] I have added a news item linked to this PR.
- [x] I have reviewed CI checks for this PR and addressed them as far as I am able.

PR #397, opened 2024-10-21
Issue #298: `epidist_prior` accepting output of `epidist_family` and improved argument documentation

### LINKED ISSUE #298: `epidist_prior` accepting a family or an `epidist_family`
In working on PR #297 I notice that `epidist_prior` currently takes a `family` as input (i.e. a default `brms` family):

```r
epidist_prior <- function(data, family, formula, prior) {
  epidist_validate(data)
  family <- brms:::validate_family(family)
  # ...
}
```

Arguably it should take an `epidist_family` as input instead (i.e. a custom `brms` family). Perhaps this would be conflating in some way the "model priors" and the "family priors". That is, by family priors are we talking about standard family priors or custom family priors?

If we did take the custom family prior then it would seem odd to dispatch on `lognormal` -- it would be dispatching on `latent_lognormal` naturally. And then we are sort of losing the shared family aspect between models. (Of course we could regex out the `lognormal` part here but then I guess for other custom families it wouldn't be obvious how to get the `lognormal` part out.)

Perhaps there is some argument again here for carrying around the `lognormal` family within the `latent_lognormal` family so that we can just keep using the custom family once we've made it.

(Some background for this is that you could argue it's ugly to have to call `family <- brms:::validate_family(family)` multiple times. It's already called in the `epidist_family` so why have to call it again in `epidist_prior` as I have done in PR #297.)

athowes:
## Description

This PR closes #298.

Changes made:

* Move lines that are called in `epidist_family_prior.class` into `epidist_family` if they are independent of `class`
* Remove adding class to object in `epidist_family` because `epidist_family` already does it
* Set `epidist_prior` to take `epidist_family` as input rather than `family`
* Alter documentation to be clearer about `family` and `epidist_family`, `formula` and `epidist_formula`, ...
* Fixes to tests...

Warrants future issues:

* I think after we get the processing in we need to go through e.g. `latent_individual.R` and fix the documentation. Right now it's a bit of a mess

## Checklist

### DIFF
--- a/R/fit.R
+++ b/R/fit.R
@@ -1,9 +1,10 @@
 #' Fit epidemiological delay distributions using a `brms` interface
 #'
-#' @inheritParams epidist_validate
-#' @param formula A formula object created using `brms::bf`. A formula must be
-#' provided for the distributional parameter `mu` common to all `brms` families.
-#' Optionally, formulas may also be provided for additional distributional
+#' @param data A `data.frame` containing line list data.
+#' @param formula An object of class [stats::formula] or [brms::brmsformula]
+#' (or one that can be coerced to those classes). A symbolic description of the
+#' model to be fitted. A formula must be provided for the distributional
+#' parameter `mu`, and may optionally be provided for other distributional
 #' parameters.
 #' @param family A description of the response distribution and link function to
 #' be used in the model. Every family function has a link argument allowing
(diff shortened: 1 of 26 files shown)


seabbs:
> I think after we get the processing in we need to go through e.g. latent_individual.R and fix the documentation. Right now it's a bit of a mess 

New issue with more details?

R/fit.R
```diff
@@ -1,9 +1,10 @@
 #' Fit epidemiological delay distributions using a `brms` interface
 #'
-#' @inheritParams epidist_validate
-#' @param formula A formula object created using `brms::bf`. A formula must be
-#' provided for the distributional parameter `mu` common to all `brms` families.
-#' Optionally, formulas may also be provided for additional distributional
+#' @param data A `data.frame` containing line list data.
+#' @param formula An object of class [stats::formula] or [brms::brmsformula]
```
seabbs, line 4: This seems like bleed through from #396

R/prior.R
```diff
@@ -12,20 +12,24 @@
 #' a warning will be shown. To prevent this warning, do not pass prior
 #' distributions for parameters which are not in the model.
 #'
-#' @param data A `data.frame` containing line list data
-#' @param family Output of a call to `brms::brmsfamily()`
-#' @param formula A formula object created using `brms::bf()`
-#' @param prior User provided prior distribution created using `brms::prior()`
+#' @inheritParams epidist
+#' @param family A description of the response distribution and link function to
```
seabbs, line 16: This seems quite imprecise and duplicative?

seabbs (approved):
LGTM

R/fit.R
athowes: No it's just bad restraint from me. I wanted to update the family documentation then it was just easy to update the documentation for other arguments of `epidist`

seabbs: Can you flag this in #396 as having happened here. Ideally avoid doing this in the future.

Issue #337, opened 2024-09-17
Use string in Ebola vignette and pass R CMD CHECK on Windows latest


athowes:
> @seabbs for some reason (!?) putting `string = "lognormal"` causes failure only in the Ebola vignette, and only on Windows latest. Going to make a new issue to resolve this but in the meantime so that we don't have failing tests on main / the rest of the branches I'm in favour of putting this in

_Originally posted by @athowes in https://github.com/epinowcast/epidist/issues/334#issuecomment-2355281427_

(This is very confusing to me, and hard to debug...)

athowes:
* Put print statements
* Try using a Windows machine
* Confirm that other string vignettes work

athowes:
The error is:

```
Error: Error: processing vignette 'ebola.Rmd' failed with diagnostics:
missing value where TRUE/FALSE needed
```

This doesn't really help but nonetheless.

athowes:
With PR #449 we reverted to using `lognormal()` hence we no longer have this issue (though we didn't solve it as is).

PR #423, opened 2024-11-06
Issue #337: Use family as string in Ebola vignette

athowes:
## Description

This is a draft PR to work on #337.

## Checklist

### DIFF
--- a/vignettes/ebola.Rmd
+++ b/vignettes/ebola.Rmd
@@ -220,7 +220,7 @@ Now we are ready to fit the latent individual model.
 
 We start by fitting a single lognormal distribution to the data. 
 This model assumes that a single distribution describes all delays in the data, regardless of the case's location, sex, or any other covariates.
-To do this, we set `formula = mu ~ 1` to place an model with only an intercept parameter (i.e. `~ 1` in R formula syntax) on the `mu` parameter of the lognormal distribution specified using `family = lognormal()`.
+To do this, we set `formula = mu ~ 1` to place an model with only an intercept parameter (i.e. `~ 1` in R formula syntax) on the `mu` parameter of the lognormal distribution specified using `family = "lognormal"`.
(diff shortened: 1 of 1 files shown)


athowes:
Some things to try here:

* Try running `ebola.Rmd` locally on a Windows machine to see if the error reproduces
*  Ensure any file paths in `ebola.Rmd` are OK on Windows
* Use `options(stringsAsFactors = FALSE)`: some R functions (especially older ones) may behave differently on Windows if stringsAsFactors is set differently
* Check for warnings suppressed on Ubuntu
* Check for differences in locale or encoding and set the locale explicitly in the R session for the vignette and specify encoding (e.g., UTF-8) in `ebola.Rmd` to prevent issues.

athowes:
@seabbs to have a look at this.

seabbs:
I don't want to be annoying and I know I drove this but I have just gone through the brms vignettes and help files and it looks like they are moving more and more towards using `lognormal()` vs a string.

Given that and given our annoying problem here shall we just return to doing so as well? 

we dropped this before because we needed to write `brms` for it to work and it was clunky. When we made the change we also suggested we could export just the families we support (as we need to write custom functionality) and I think this does make sense.

seabbs:
We would do this using this approach and enforce all new models to hit all currently supported families: https://forum.posit.co/t/best-practices-for-documenting-re-exports/89254

seabbs:
If we agree we want to do this I am happy to make the code changes ahead of the release next week (also happy for @athowes to cover)

athowes:
I'm fine with using `brms::lognormal()` if we can't resolve this bug.

seabbs:
So I am saying we reexport and so don't use brms:: just lognormal()

athowes:
Yes I am on the same page. Was just using `brms::lognormal()` to clarify which function rather than that we explicitly call it (but can see how that would have been confusing).

PR #434, opened 2024-11-14
Issue #432: Remove exported filtering functions

### LINKED ISSUE #432: Remove `filter_obs()` (if possible)
Like #431 I think we want to get rid of we can.

athowes:
## Description

This PR closes #432.

It moves one filtering function into the two places it's used. And deletes the other which is used no where.

We could have follow up issue on adding back in some better simulation code about this, but we could also do that later.

## Checklist


vignettes/approx-inference.Rmd
```diff
@@ -110,6 +110,15 @@ sdlog <- 0.5
 obs_time <- 25
 sample_size <- 200
 
+filter_obs_by_obs_time <- function(linelist, obs_time) {
```
seabbs, line 113: I really don't like putting function definitions into the vignettes like this. I think all we need is the filter line so can you just use that as needed vs this approach

athowes: Sure I agree that is probably better. This is quicker to do now and could be quicker when we do the refactor to a future function

athowes:
@seabbs lmk if you want me to take back over on getting this passing tests

seabbs (approved):
I think this looks okay now

PR #447, opened 2024-11-16
Issue #398: Make documentation consistent across models

### LINKED ISSUE #398: Update documentation for specific models
After we have the preprocessing interface solidified, we should go back and make sure that the models we have are documented well taking that preprocessing into account.

For example right now:

```r
#' Prepare latent individual model
#'
#' @param data A `data.frame` containing line list data
#' @family latent_individual
#' @export
as_latent_individual <- function(data) {
  UseMethod("as_latent_individual")
}
```

I was going to replace `data` here with an `@inheritParam` from somehwere, but really this `data` should be a different kind of parameter. One that is before the processing. Perhaps it should say "an object created using `as_linelist()` (or another constructor).

Another example issue:

```r
#' Define the model-specific component of an `epidist` custom formula
#'
#' @param data A `data.frame` containing line list data
#' @param formula As produced by [brms::brmsformula()]
#' @param ... ...
#' @method epidist_formula_model epidist_latent_individual
#' @family latent_individual
#' @export
epidist_formula_model.epidist_latent_individual <- function(
  data, formula, ...
) {
  # data is only used to dispatch on
  formula <- stats::update(
    formula, delay | vreal(relative_obs_time, pwindow, swindow) ~ .
  )
  return(formula)
}
```

This shouldn't say `formula` as produced by `brms::brmsformula()` since it needs to have been run through `epidist_formula`. How to do this and share the argument nicely?

We also need to do this for the direct / pass through model of PR #393.

athowes:
## Description

This PR closes #398.

* Removes documentation mentions of "A `data.frame` containing line list data." where it's not about the `epidist_linelist_data` class
* Trys to more generally describe the inputs to each function
* Moves a function in the `naive_model.R` position to be standard with `latent_model.R`

Note that I don't touch some mentions of "direct" in this PR which are handled by #446.

This PR raised some follow up issues which I have added (https://github.com/epinowcast/epidist/issues/448).

## Checklist


R/epidist.R
```diff
@@ -1,6 +1,8 @@
 #' Fit epidemiological delay distributions using a `brms` interface
 #'
-#' @param data A `data.frame` containing line list data.
+#' @param data An object with class corresponding to an implemented model.
+#' The currently available models classes are `epidist_latent_model` and
```
seabbs, line 4: I'd suggest taking this out and instead linking to somewhere else as it will be easy to get outdated

seabbs (dismissed):

R/epidist.R
athowes: What do you think of this idea https://github.com/epinowcast/epidist/issues/448?

athowes: [dd6e22c](https://github.com/epinowcast/epidist/pull/447/commits/dd6e22c8548ca8836518e0d645f7b884a9961597)

seabbs (approved):

PR #449, opened 2024-11-16
Issue #438: Use class as default for family and reexport common families and formula function

### LINKED ISSUE #438: Revert to using family as family class rather than string
As discussed in #423 comments.

What needs to be done:

* Revert to family as e.g. `lognormal()` in vignettes
* And in default to function
* And in test suite
* Add back in tests that the string version works
* Export common families so that `brms::` isn't needed

This will fix #332 also.

athowes:
## Description

This PR closes #438.

Reexports `brms::lognormal`, `brms::weibull`, `stats::Gamma` and `brms::bf`.

Changes default to `lognormal()` away from `"lognormal"`.

I've also altered the test on the recovery of simulated `fit_sex` parameters to use the right link functions so that they pass. Also on sampling from prior.

I am wondering if `lognormal()` has different link functions by default than `"lognormal"`. In particular it uses `exp()` on the SD whereas maybe the string one just uses identity (which would be bad to do, so good we've moved away from it).

## Checklist


athowes:
These are stochastic tests failing. Wonder if they are real fails or not.

seabbs (approved):

PR #459, opened 2024-11-20
Issue #442: Generalise to all brms distributions

### LINKED ISSUE #442: Make posterior_predict_latent etc. use function factories
rather than requiring a function per distribution as we do now. At the same time we should update to use `primarycensored` vs the current custom functionality.

seabbs:
## Description

This PR closes #442. It adds function factories and uses `primarycensored` to generalise the latent model to all `brms` supported distributions.

My outstanding uncertainty here is the naming and organisation of these functions. It would also be nice to still support custom families but I think that might need to be a new issue/stretch goal (as is quite edge case functionality anyway).

## Checklist


tests/testthat/test-gen.R
```diff
@@ -0,0 +1,120 @@
...
+  expect_gt(max(q_gamma), 0.9)
+  expect_lt(mean(q_gamma), 0.65)
+  expect_gt(mean(q_gamma), 0.35)
+})
+
+test_that("epidist_gen_posterior_epred returns a function that creates arrays with correct dimensions", { # nolint: line_length_linter.
+  skip_on_cran()
+  # Test lognormal
+  epred <- prep_obs |>
+    tidybayes::add_epred_draws(fit)
```
seabbs, line 108: this fails if you do it directly using the generated function and not via tidybayes. I can't work out why but I also can't find a real use case it impacts yet

vignettes/ebola.Rmd
```diff
@@ -317,7 +320,6 @@ Figure \@ref(fig:epred)B illustrates the higher mean of men as compared with wom
 
 ```{r}
 epred_draws <- obs_prep |>
-  as.data.frame() |>
```
seabbs, line 320: I removed these as it doesn't look like they are doing anything right now

vignettes/epidist.Rmd
```diff
@@ -297,38 +299,57 @@ data.frame(
   tab_caption("(ref:pars)")
 ```
 
+For example, we can use the built in `summary()` function
+
+```{r}
```
seabbs, line 304: I added this originally to sense check the changes but I also quite like showing this early

vignettes/epidist.Rmd
```diff
@@ -297,38 +299,57 @@ data.frame(
...
-  geom_function(
-    data = data.frame(x = c(0, 30)),
-    aes(x = x), fun = dlnorm,
-    args = list(
-      meanlog = mean(pred$mu),
-      sdlog = mean(pred$sigma)
-    ),
-    col = "#CC79A7"
-  ) +
+  # Plot 100 sampled posterior distributions
```
seabbs, line 339: This was also a sense check but I've left it in as I think it might be useful

seabbs:
@kylieainslie I think this PR has put us in a place where we can think about testing mixtures using the brms `mixture()` family (which I think should now "just work" with the double censoring and truncation methods in the latent model generalised here

seabbs:
The coverage gap we see here is not a new gap but showing up due to copying the function into a new location. In general the log likelihood tests are a bit weak but its also really hard to test. A potentially sensible candidate for a mocked test sometimes in the future

R/gen.R
```diff
@@ -0,0 +1,73 @@
+#' Create a function to draw from the posterior predictive distribution for a
+#' latent model
+#'
+#' This function creates a function that draws from the posterior predictive
+#' distribution for a latent model using [primarycensored::rpcens()] to handle
+#' censoring and truncation. The returned function takes a prep argument from
+#' `brms` and returns posterior predictions. This is used internally by
+#' [brms::posterior_predict()] to generate predictions for latent models.
+#'
+#' @inheritParams epidist_family_model
```
athowes, line 10: The documentation here you get is:

> Output of a call to [brms::brmsfamily()](https://paulbuerkner.com/brms/reference/brmsfamily.html) with additional information as provided by [.add_dpar_info()](file:///Users/adamhowes/Downloads/website/reference/dot-add_dpar_info.html)

Is this the intention? Is the additional `.add_dpar_info()` needed?

R/gen.R
```diff
@@ -0,0 +1,73 @@
...
+#' * `vreal1`: relative observation time
+#' * `vreal2`: primary event window
+#' * `vreal3`: secondary event window
+#'
+#' @seealso [brms::posterior_predict()] for details on how this is used within
+#' `brms`, [primarycensored::rpcens()] for details on the censoring approach
+#' @autoglobal
+#' @family gen
+#' @export
+epidist_gen_posterior_predict <- function(family) {
```
athowes, line 24: Is this function specific to the latent model? If so I'd consider including it in the name

R/gen.R
```diff
@@ -0,0 +1,73 @@
+#' Create a function to draw from the posterior predictive distribution for a
+#' latent model
+#'
+#' This function creates a function that draws from the posterior predictive
+#' distribution for a latent model using [primarycensored::rpcens()] to handle
+#' censoring and truncation. The returned function takes a prep argument from
```
athowes, line 6:
```suggestion
#' censoring and truncation. The returned function takes a `prep` argument from
```

R/gen.R
```diff
@@ -0,0 +1,73 @@
...
+#' `brms` and returns posterior predictions. This is used internally by
+#' [brms::posterior_predict()] to generate predictions for latent models.
+#'
+#' @inheritParams epidist_family_model
+#'
+#' @return A function that takes a prep argument from brms and returns a matrix
```
athowes, line 12:
```suggestion
#' @return A function that takes a `prep` argument from `brms` and returns a matrix
```

R/gen.R
```diff
@@ -0,0 +1,73 @@
...
+#' of posterior predictions, with one row per posterior draw and one column
+#' per observation. The prep object must have the following variables:
```
athowes, line 14:
```suggestion
#' per observation. The `prep` object must have the following variables:
```

athowes: And more later down

R/gen.R
athowes: I can see that the `log_lik` one has the name in. I don't think these functions will work by default with e.g. the marginal model. Is the point that they'll be modified then? Some question about whether for the time being for consistency they should be named `latent` given that they don't work with the marginal model as is (and the marginal model doesn't exist yet)?

R/latent_model.R
```diff
@@ -89,12 +89,76 @@ epidist_family_model.epidist_latent_model <- function(
...
+#' * `vreal2`: primary event window
+#' * `vreal3`: secondary event window
+#'
+#' @family latent_model
+#' @autoglobal
+epidist_gen_log_lik_latent <- function(family) {
+  # Get internal brms log_lik function
+  log_lik_brms <- .get_brms_fn("log_lik", family)
+
+  .log_lik <- function(i, prep) {
```
athowes, line 126: So I guess it's not possible to replace this part with calls to `primarycensored` functions?

R/utils.R
```diff
@@ -157,3 +157,22 @@
...
+#' Get a brms function by prefix and family
+#'
+#' Helper function to get internal brms functions by constructing their name
+#' from a prefix and family. Used to get functions like `log_lik_*`,
+#' `posterior_predict_*` etc.
+#'
+#' @param prefix Character string prefix of the brms function to get (e.g.
+#' "log_lik")
+#'
+#' @param family A brms family object
```
athowes, line 170: Reuse documentation here?

_pkgdown.yml
```diff
@@ -69,6 +69,10 @@ reference:
   desc: Specific methods for the naive model
   contents:
   - has_concept("naive_model")
+- title: Generator model functions
+  desc: Generator functions for creating family specific model utilities.
```
athowes, line 73:
```suggestion
  desc: Generator functions for creating family specific model utilities
```

tests/testthat/test-gen.R
athowes: As in if users try using the generated functions themselves somehow they will run into errors? I guess that's non-ideal if so. Maybe we want unit tests on the generated functions. If not solved here I think worth follow up issue

vignettes/epidist.Rmd
```diff
@@ -297,38 +299,57 @@ data.frame(
   tab_caption("(ref:pars)")
 ```
 
+For example, we can use the built in `summary()` function
+
+```{r}
+summary(fit)
+```
+
+to see that the model has converged and recovered the delay distribution distribution paramters reasonably well.
```
athowes, line 308:
```suggestion
to see that the model has converged and recovered the delay distribution distribution parameters reasonably well.
```

vignettes/epidist.Rmd
athowes: Yeah I think it's better, thanks. IMO the black line is too thick -- maybe 1.5 / I would try to distinguish in some other way but it's minor

vignettes/ebola.Rmd
athowes, line 1: The `ebola.Rmd` file also talks about the `posterior_epred_latent_lognormal` function specifically. Suggest find any instances of replaced functions in package and rewrite so that the text makes sense

tests/testthat/test-gen.R
seabbs: it looks like it was also a problem with the previous implementations so its not clear to me what is going on but agree will make a new issue

R/gen.R
seabbs: as discussed the idea is that some are generic to the data generating process, some are just generic (i.e the epred approach) and some are model specific (i.e this latent one). We could just stick them all in the model for but I would prefer to aim to generalise from the get go.

seabbs: I'll make an issue to explore how generalisable these are

R/gen.R
```diff
@@ -0,0 +1,73 @@
...
+epidist_gen_posterior_predict <- function(family) {
+  dist_fn <- .get_brms_fn("posterior_predict", family)
+
+  rdist <- function(n, i, prep, ...) {
+    prep$ndraws <- n
+    do.call(dist_fn, list(i = i, prep = prep))
+  }
+
+  .predict <- function(i, prep, ...) {
+    relative_obs_time <- prep$data$vreal1[i]
```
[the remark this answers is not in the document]
seabbs: closing in favour of the duplicate comment

R/latent_model.R
seabbs: no because this isn't the primarycensored model / likelihood

R/gen.R
seabbs: See #462

tests/testthat/test-gen.R
seabbs: See #463

athowes (approved):
🥳

PR #426, opened 2024-11-07
Issue #221: Add marginal model

athowes:
## Description

This PR implements the marginal model as in #221.

* The file `marginal_model-scratch.R` implements a prototype outside of functions
* This prototype is implemented into the standard structure
* Uses `primarycensored` function `primarycensored_lpmf` which is passed into via a wrapper in `marginal_model/functions.stan`
  * Uses only the analytic versions of this distribution
* Adds tests for new `marginal_model`

Possible steps forward here:

* But uses non vectorised version of LPMF: want to extend to be using vectorised version of LPMF
* Could want to extend to being able to use non analytic solution in a natural way

## Checklist

- [x] My PR is based on a package issue and I have explicitly linked it.
- [x] I have included the target issue or issues in the PR title in the for Issue(s) *issue-numbers*: PR title
- [x] I have read the [contribution guidelines](https://github.com/epinowcast/.github/blob/main/CONTRIBUTING.md).
- [x] I have tested my changes locally.
- [x] I have added or updated unit tests where necessary.
- [x] I have updated the documentation if required.
- [x] My code follows the established coding standards.
- [x] I have added a news item linked to this PR.
- [x] I have reviewed CI checks for this PR and addressed them as far as I am able.

Issue #420, opened 2024-11-05
Tracking unexported `brms` functions


athowes:
We use four `brms` functions:

* `brms:::dpar_bounds`
* `brms:::validate_data`
* `brms:::validate_family`
* `brms:::validate_formula`

See:

```
N  checking dependencies in R code (2.2s)
   Unexported objects imported by ':::' calls:
     ‘brms:::dpar_bounds’ ‘brms:::validate_data’ ‘brms:::validate_family’
     ‘brms:::validate_formula’
     See the note in ?`:::` about the use of this operator.
```

This issue is a place to track progress on removing dependence on these functions, or them being exposed within `brms`. Adam to link issue raised in `brms` repository about this.

athowes:
Adam to link issue in `brms`.

athowes:
https://github.com/paul-buerkner/brms/issues/1676

seabbs:
brms:::log_lik_weight and all internal family log_like and posterior predict functions

athowes:
Note that this issue blocks CRAN submission.

PR #474, opened 2024-11-24
Issue #267: Refactor to allow custom event priors and marginalise the latent likelihood

### LINKED ISSUE #267: Make `log_lik` methods correct by obtaining latent parameter draws
In PR #259 we added a slightly incorrect version of the `log_lik_latent_lognormal` function. This allowed progress on model comparison in the Ebola vignette (#48).

It would be good to make this correct however.

For background see:

* #240
* A [thread](https://discourse.mc-stan.org/t/log-likelihood-for-brms-custom-family-that-depends-on-latent-variables/36415) on the Stan forum summarising the issue

In short I see the two most promissing routes as:

1. Try to put the required `pwindow_raw` and `swindow_raw` draws into some part of the `prep` object
2. Use `...` to pass in the `fit` object then extract the draws

Future investigation required to figure out how to do this!

seabbs:
## Description

This PR closes #267 by moving to a formula approach for `pwindow` and `swindow`. On top of allowing for the correct likelihood this also allows users to set priors on pwindow and swindow. This could be extended to allow for custom formulas (for pwindow this would be a useful feature but not for swindow).

## Checklist

- [x] I have added a news item linked to this PR.


R/latent_model.R
```diff
@@ -83,12 +83,12 @@ epidist_family_model.epidist_latent_model <- function(
   # Really the name and vars are the "model-specific" parts here
   custom_family <- brms::custom_family(
     paste0("latent_", family$family),
-    dpars = family$dpars,
-    links = c(family$link, family$other_links),
-    lb = c(NA, as.numeric(lapply(family$other_bounds, "[[", "lb"))),
-    ub = c(NA, as.numeric(lapply(family$other_bounds, "[[", "ub"))),
+    dpars = c(family$dpars, "pwindow", "swindow"),
```
athowes, line 86: Am I right to think that placing parameters into `dpars` rather than injecting them via Stan shouldn't have any implications for the computational efficiency of the model? At the end of the day it's the same Stan code?

R/latent_model.R
```diff
@@ -168,13 +165,45 @@ epidist_gen_log_lik_latent <- function(family) {
...
+  # Update main formula
   formula <- stats::update(
     formula, delay | vreal(relative_obs_time, pwindow, swindow) ~ .
   )
+
+  # Only update pwindow/swindow formulas if intercept only
+  fixed_dpars <- names(formula$pfix)
+  formula_dpars <- names(formula$pforms)
+
+  # Check if pwindow needs updating
```
athowes, line 177: Is the point of this to make sure that the formula for `pwindow` and `swindow` are basically random effects on `.row_id`? Do we cover all cases so that someone can't set another formula? Or as it looks like now if someone puts something other than `pwindow ~ 1` then that will be kept? Are there use cases for something other than `pwindow ~ 0 + as.factor(.row_id)`?

seabbs:
Noting this has been slowed down as I realised that reparameterisation wasn't working currently with general families and `.replace_prior` didn't appear to work as expected when used in `epidist_prior`.

R/latent_model.R
seabbs: yes

seabbs:
I don't think that matrix multiplication issue is possible to get around so I think I would argue this PR is repurposed to address the issues found whilst exploring this and we hit pause on custom priors and look for another method for correcting the likelihood

R/epidist.R
```diff
@@ -15,6 +15,10 @@
 #' @param prior One or more `brmsprior` objects created by [brms::set_prior()]
 #' or related functions. These priors are passed to [epidist_prior()] in the
 #' `prior` argument.
+#' @param merge_priors If `TRUE` then merge new priors with existing ones, if
```
athowes, line 18: Suggest to talk about the priors provided by the user rather than "new priors" and to be clearer on what "existing ones" are. As in maybe they don't know that we try to set our own defaults.

vignettes/faq.Rmd
```diff
@@ -181,7 +181,7 @@ quantile(pred$sd, c(0.01, 0.1, 0.25, 0.5, 0.75, 0.9, 0.99))
 
 We recommend use of the [`priorsense`](https://github.com/n-kall/priorsense) package [@kallioinen2024detecting] to check how sensitive the posterior distribution is to perturbations of the prior distribution and likelihood using power-scaling analysis:
 
-```{r}
+```{r, eval = FALSE}
```
athowes, line 184: Is this intentional? It won't produce the figure then?

R/latent_model.R
```diff
@@ -168,13 +112,36 @@ epidist_gen_log_lik_latent <- function(family) {
 #' @export
 epidist_formula_model.epidist_latent_model <- function(
     data, formula, ...) {
-  # data is only used to dispatch on
+  # Update main formula
   formula <- stats::update(
     formula, delay | vreal(relative_obs_time, pwindow, swindow) ~ .
   )
```
athowes, line 118: At one point didn't we have a formula on `pwindow` and `swindow`?

athowes (commented):
* I think the way to do the reparameterisation is cool. I do worry about it being a bit hacky, and relying on regex. Perhaps though the regex is relatively robust and the pattern it's looking for in `brms` Stan code can be relied not to change. Also to note I'm unsure about why this feature is a part of this PR. (Is it related to refactor to use formula for `pwindow` and `swindow`?). Also if it's not using S3 suggest dropping S3.
* On the `log_lik`, `posterior_predict` and `posterior_epred` I think the generalisation beyond being about the latent model is good. I wonder about how we will extend this to working with the `marginal_model` (PR #426). For the naive model it already has these methods as it doesn't need a custom family. One could wonder about simulating from the right likelihood after using the naive model, but perhaps it's not going to happen / be a useful feature.
* It's exciting to have priors on the `pwindow_raw` and `swindow_raw`. I wonder about how easy it will be for users on the journey to changing those priors (knowing which parameters to set).
* About the prior infrastructure, it seems to be getting quite complicated and or brittle. Perhaps you disagree. I wonder about what we might be able to do to simplify it. I think at some level I feel quite bad about including an argument like `merge_priors` in the main `epidist()` function as it shows our approach isn't really working that well. I wonder what use cases the "built in approach isn't flexible enough for" -- is it mainly this one about prior for all coefficients? I was a bit confused by that because in the previous approach I thought we were matching on all entries in the dataframe. Now it seems like we are doing some regex and it's a different approach.
* Is there some way we can document how users can make use of the priors on window functionality? And when they would do it? Inclusion in some vignette? Create new issue on this?

R/family.R
```diff
@@ -43,29 +43,66 @@ epidist_family_model.default <- function(data, family, ...) {
 #' Reparameterise an `epidist` family to align `brms` and Stan
 #'
 #' @inheritParams epidist_family
-#' @rdname epidist_family_reparam
+#' @rdname epidist_family_param
 #' @family family
 #' @export
-epidist_family_reparam <- function(family, ...) {
-  UseMethod("epidist_family_reparam")
+epidist_family_param <- function(family, ...) {
```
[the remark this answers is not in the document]
seabbs: I left this as it is in case the default parame strategy doesn't work in all cases (i.e you can write a custom workaround). I thought we would monitor this and take it out if it appears to work well.

R/family.R
```diff
@@ -43,29 +43,66 @@ epidist_family_model.default <- function(data, family, ...) {
...
-#' @family family
-#' @export
-epidist_family_reparam.default <- function(family, ...) {
-  family$reparam <- family$dpars
-  return(family)
-}
-
-#' Reparameterisation for the gamma family
+#' This function extracts the Stan parameterisation for a given brms family by
+#' creating a dummy model and parsing its Stan code. It looks for the log
```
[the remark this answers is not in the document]
seabbs: good point. This feels like a new issue to track/update various uses

R/family.R
```diff
@@ -43,29 +43,66 @@ epidist_family_model.default <- function(data, family, ...) {
...
+  dummy_mdl <- make_stancode(y ~ 1, data = df, family = class(family)[1])
+
+  # get the lowered family name
+  family_name <- tolower(class(family)[1])
+
+  # Extract the Stan parameterisation from the dummy model code
+  lpdf_pattern <- paste0(
+    "target \\+= ", family_name, "_(lpdf|lpmf)\\(Y \\| ([^)]+)\\)" # nolint
+  )
+  lpdf_match <- regexpr(lpdf_pattern, dummy_mdl)
```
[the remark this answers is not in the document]
seabbs: Good point - I think we also need to look for sigma

vignettes/faq.Rmd
seabbs: yes - see the comment in the PR issue it errors with the new likelihood

R/latent_model.R
```diff
@@ -227,36 +190,24 @@ epidist_stancode.epidist_latent_model <- function(
...
-      block = "data",
-      scode = "array[N - wN] int noverlap;",
-      x = filter(data, woverlap == 0)$.row_id,
-      name = "noverlap"
-    ) +
     stanvar(
       block = "data",
       scode = "array[wN] int woverlap;",
-      x = filter(data, woverlap > 0)$.row_id,
+      x = as.integer(filter(data, woverlap > 0)$.row_id),
```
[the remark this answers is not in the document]
seabbs: this is just to enforce it so it doesn't erorr with stan we can drop

R/latent_model.R
seabbs: yes see the comment in the PR about this not being scalable due to the matrix multiplication we can't avoid

R/family.R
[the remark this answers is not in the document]
seabbs: #480

R/gen.R
```diff
@@ -1,5 +1,75 @@
...
+        i = i,
+        prep = prep,
+        pwindow = pwindow,
+        swindow = swindow,
+        D = relative_obs_time,
+        dprimary = stats::dunif,
+        log = TRUE
+      )
+    })
+    lpdf <- brms:::log_lik_weight(lpdf, i = i, prep = prep) # nolint
```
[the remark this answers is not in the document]
seabbs: https://github.com/epinowcast/epidist/issues/420#issuecomment-2506478324

athowes (approved):

PR #513, opened 2025-01-28
Issue #508: Aggregate data class

seabbs:
## Description

This PR closes #508 and #412 by adding an aggregate data class and finalising support in `as_epidist_marginal_model`.

## Checklist

- [x] My PR is based on a package issue and I have explicitly linked it.
- [x] I have included the target issue or issues in the PR title in the for Issue(s) *issue-numbers*: PR title
- [x] I have read the [contribution guidelines](https://github.com/epinowcast/.github/blob/main/CONTRIBUTING.md).
- [x] I have tested my changes locally.
- [x] I have added or updated unit tests where necessary.
- [x] I have updated the documentation if required.
- [x] My code follows the established coding standards.
- [x] I have added a news item linked to this PR.
- [x] I have reviewed CI checks for this PR and addressed them as far as I am able.

PR #515, opened 2025-01-29
update marginal model docs

### LINKED ISSUE #512: Clarification on use of `weight` argument in `as_epidist_marginal_model`
PR #509 adds support for users providing a column which contains counts / weights. Docs are:

> A column name to use for weighting the data in the likelihood. Default is NULL. Internally this is used to define the 'n' column of the returned object.

Compatible with `epidist_linelist_data` (which doesn't naturally support weighted linelist / aggregate, hence I suppose https://github.com/epinowcast/epidist/issues/508).

Some notes:

* We had back and forth on design suggesting that aggregation needed to be internal to the marginal model. I guess that isn't a blocker because aggregation still occurs in the marginal model, but there is no harm in having this additional aggregatation. Essentially it's the user saying "all this data is the same" which can never be non-optimal (unless they are wrong). This then gets further aggregated up if it's possible to.
* Likely want examples demonstrating this feature e.g. in vignette

So I guess scope of this issue could add docs to note that the marginal model will perform further aggregation if it is possible to given the formula etc. that you specify. Or that could be best left to longer form vignette.

kgostic:
## Description

This PR closes #512.

- Added @details and additional documentation to explain internal aggregation, and the weights option in the marginal model.

Reviewer notes:
1. I did my best to follow the set of variables needed for grouping based on the [code](https://github.com/epinowcast/epidist/blob/f66f29b1b86cf7928f157701b9d0c953d000c5bc/R/marginal_model.R#L172), but please check that I got this right @seabbs! 
2. Currently, this documentation only renders for `?as_epidist_marginal_model.epidist_linelist_data`, whereas it would probably be more useful if it rendered for `?as_epidist_marginal_model`. The docs for the latter are so sparse as to not really be useful, but that's all people will know to search for. I wasn't sure how to change this, or whether changing it would have unintended consequences. Please weigh in!

## Checklist

- [x ] My PR is based on a package issue and I have explicitly linked it.
- [x ] I have included the target issue or issues in the PR title in the for Issue(s) *issue-numbers*: PR title
- [x ] I have read the [contribution guidelines](https://github.com/epinowcast/.github/blob/main/CONTRIBUTING.md).
- [x ] I have added or updated unit tests where necessary.
- [x ] I have updated the documentation if required.

### DIFF
--- a/R/marginal_model.R
+++ b/R/marginal_model.R
@@ -18,8 +18,43 @@ as_epidist_marginal_model <- function(data, ...) {
 #'   Default is 2.
 #' @param weight A column name to use for weighting the data in the
 #'   likelihood. Default is NULL. Internally this is used to define the 'n'
-#'   column of the returned object.
+#'   column of the returned object. See details.
 #' @param ... Not used in this method.
+#' @details To ensure efficient computation, the model automatically
+#' identifies groups of individuals in the data that contribute identically
+#' to the likelihood, and evaluates the likelihood at the group level. Groups
+#' are defined by unique combinations of: {ptime_upr, stime_upr, stime_lwr,
+#' relative_obs_time, pwindow, swindow, and other_vars}, where the first six
+#' fields are defined by `as_epidist_linelist_data()` (see also the [getting started vignette](linelihttps://epidist.epinowcast.org/dev/articles/epidist.html#data)),
+#' and where `other_vars` includes other variables used in the model forumla,
+#' (e.g. age, sex, or location).
+#'
+#' The `weight` option can be used for convenience when working with data that
+#' contain too many individuals to represent easily in linelist format. For
+#' example, `prepped` and `prepped_weighted` would be
+#' interpreted identically by epidist():
+#'
+#' ```
+#' ## Prep as a linelist
+#' prepped <- linelist |>
+#'   as_epidist_linelist_data() |>
+#'   as_epidist_marginal_model(linelist)
+#' ```
+#'
+#' or
+#'
+#' ```
+#' ## Aggregate by relevant variables
+#' prepped_weighted <- linelist |>
+#'  as_epidist_linelist_data() |>
+#'  # summarise by unique combinations of variables
+#'   group_by(ptime_upr, stime_upr, stime_lwr,
+#'            relative_obs_time, pwindow, swindow,
+#'            other_vars) |>
+#' summarise(n = n())
+#' # prep with weight option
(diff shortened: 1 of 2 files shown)


kgostic:
Keeping this as a draft, but I think I need some feedback before I can do anything else. As far as I can tell the CI issues are being resolved elsewhere and unrelated to the changes made here.

seabbs:
Thanks for this @kgostic.

>  did my best to follow the set of variables needed for grouping based on the

I think the aggregate data class may help resolve these issues.

> There wasn't a pre-built dataset that was obviously good to use in the example, so I wrote something hypothetical and put into https://github.com/details?rgh-link-date=2025-01-29T23%3A04%3A03Z instead of https://github.com/examples?rgh-link-date=2025-01-29T23%3A04%3A03Z. Hope that's ok.

I have been resolving this in #513 as I also could make examples without one there. It probably makes sense to adopt that approach. I think there is also some context etc in that PR that might help here and so perhaps it is worth waiting on that and then refactoring/reviewing this. 


> Currently, this documentation only renders for ?as_epidist_marginal_model.epidist_linelist_data, whereas it would probably be more useful if it rendered for ?as_epidist_marginal_model. The docs for the latter are so sparse as to not really be useful, but that's all people will know to search for. I wasn't sure how to change this, or whether changing it would have unintended consequences. Please weigh in!

I think this should be a more general issue as it impacts all the s3 methods (there may in fact be one as its been on the to do list for a while). There are two ways to go with this 1. As you suggest put everything on the same doc page and 2. Use a linking system (as we do for model/data related functions at the moment so people can navigate. epinowcast uses the latter as it extends base s3 generics and scoringutils uses the former. For scoringutils that has been a problem as that central doc page became so full (due to many methods) it became hard to read and so over time we have been moving from 1. to 2.

seabbs:
Touching base one this

kgostic:
It sounds like you added some data for an example in #513 and that changing the docs so that the help files are visible is a separate issue?

I'm not sure how quickly I'll get to this but can try to update the example bit later this week. I think linking to the help docs is pretty important.

seabbs:
> changing the docs so that the help files are visible is a separate issue?

I added more information and links in that information between functions but I didn't change the approach to s3 documentation. As above I think that this is a separate issue really as from past experience there are a few things to consider and a few viable approaches. 

Noted on the update.

seabbs:
Closing for now as I think this is stuck. Happy to reopen or have further discussion in the issues.

